PR #20, opened 2022-09-14
[RPA-426] Validate process name during creation of a new process

kzuraniewski:


runbotics-orchestrator/src/main/java/com/runbotics/web/rest/ProcessResource.java
```diff
@@ -258,6 +259,27 @@ public ResponseEntity<ProcessDTO> getProcess(@PathVariable Long id) {
...
+     * @return the {@link ResponseEntity} with status {@code 200 (OK)} or with status {@code 409 (Conflict)}.
+     */
+    @PreAuthorize("@securityService.checkFeatureKeyAccess('" + FeatureKeyConstants.PROCESS_READ + "')")
+    @GetMapping("/processes/name/{processName}/is-available")
+    public ResponseEntity<Void> checkIfProcessExists(@PathVariable String processName) {
+        log.debug("REST request to check if Process of given name is available : {}", processName);
+        Optional<ProcessDTO> processDTO = processService.findByName(processName);
+        
+        if (processDTO.isPresent()) {
+            log.debug("Process of name {} is not available", processName);
```
preludium: it should be more like "Name of the process {} is not available"

in addition add prefix for logs `=>` and `<=` whether it is log at the start of the endpoint or log before the end of processing

runbotics/runbotics-orchestrator-ui/src/store/slices/Process/Process.thunks.ts
```diff
@@ -103,3 +103,8 @@ export const deleteProcess = createAsyncThunk<void, { processId: number }>(
         await Axios.delete(`/api/processes/${processId}`);
     },
 );
+
+export const isProcessAvailable = createAsyncThunk<void, { processName: string }>(
+    'processes/name/is-available',
+    async ({ processName }) => Axios.get<void>(`/api/processes/name/${processName}/is-available`).then(response => response.data),
```
preludium: put a new line every chain (before then method) to improve readability

preludium: I think that we dont need this then clause anyway, because there will be no response body

runbotics/runbotics-orchestrator-ui/src/translations/pl/process.json
```diff
@@ -35,6 +35,8 @@
     "Process.Add.Title": "Add new process",
     "Process.Add.Form.Fields.Name": "Name",
 
+    "Process.Error.NameNotAvailable": "This name is already taken",
```
preludium: provide correct polish translation

runbotics/runbotics-orchestrator-ui/src/views/process/AddProcess.tsx
```diff
@@ -34,6 +36,14 @@ const formData: IProcess = {
     definition: emptyBpmn,
 };
 
+enum ErrorType {
+    NAME_NOT_AVAILABLE,
```
preludium: if you dont provide value for enum, by the default it has assigned 0, 1, 2 and so on.
lets put there string value, it can be the same as the key

runbotics/runbotics-orchestrator-ui/src/views/process/AddProcess.tsx
```diff
@@ -34,6 +36,14 @@ const formData: IProcess = {
...
+}
+
+const errorMessages: Record<ErrorType, string> = {
+    [ErrorType.NAME_NOT_AVAILABLE]: translate('Process.Error.NameNotAvailable'),
```
preludium: as mentioned above `ErrorType.NAME_NOT_AVAILABLE` returns numeric value which is not correct for js objects (you cannot do it directly if you create new object

runbotics/runbotics-orchestrator-ui/src/views/process/AddProcess.tsx
```diff
@@ -100,16 +132,16 @@ const AddProcess = () => {
...
                     <SvgIcon fontSize="small">
                         <PlusIcon />
                     </SvgIcon>
-                )}
+                }
             >
                 {translate('Process.Add.ActionName')}
             </Button>
-            {show && <AddProcessDialog open={show} onClose={() => setShow(false)} onAdd={handleAdd} />}
+            {showDialog && <AddProcessDialog open={showDialog} onClose={() => setShowDialog(false)} onAdd={handleAdd} />}
```
preludium: i know that it was already there but lets correct it
I mean that we decide about visibility of this dialog by open prop and `{showDialog && <AddProcessDialog...`

IMO use only open prop, check if it works correctly otherwise adjust logic in this component

runbotics/runbotics-orchestrator-ui/src/views/process/AddProcess.tsx
```diff
@@ -43,18 +53,40 @@ type AddProcessDialogProps = {
...
+            setErrorType(ErrorType.NAME_NOT_AVAILABLE);
+            return;
+        }
+        setErrorType(null);
+
         const result = await Axios.post<IProcess>('/api/processes', e.formData);
         props.onAdd(result.data);
     };
 
+    const extraErrors = errorType !== null ? {
```
preludium: send me a screen how this message is displayed on modal

preludium (changes requested):

preludium (approved):

PR #36, opened 2022-09-30
Attended process validation

FranekMarciniak:


runbotics/runbotics-scheduler/src/scheduler/scheduler.service.ts
```diff
@@ -79,23 +79,35 @@ export class SchedulerService implements OnApplicationBootstrap {
...
+
+      const emptyVariables = requiredVariables.map((variable) => {
+        if (isObject(passedVariables[variable]) || passedVariables[variable] !== '') {
+          return null;
+        }
+        return variable;
+      }).filter((variable) => variable !== null);
+
+      if (emptyVariables.length > 0) {
+        const err = { message: `You haven't provided variables for attended process: ${emptyVariables.join(', ')}`, statusCode: 400 };
```
preludium: just pass error message, status code is already covered by specific `BadRequestException`

preludium (dismissed):

runbotics/runbotics-scheduler/src/scheduler/scheduler.service.ts
FranekMarciniak: tru

preludium (approved):

PR #155, opened 2023-03-02
fix starting process via schedule

preludium:


runbotics/runbotics-scheduler/src/queue/queue.service.ts
```diff
@@ -124,7 +124,7 @@ export class QueueService implements OnModuleInit {
     }
 
     async validateProcessAccess({ process, user, triggered = false }: ValidateProcessAccessProps) {
-        const hasAccess = (process.createdBy.id === user?.id);
+        const hasAccess = process.createdBy?.id === user?.id;
```
star-in-the-jar, line 127: I hope you considered that

```
> undefined === undefined
< true
```

runbotics/runbotics-scheduler/src/queue/schedule-process/schedule-process.controller.ts
```diff
@@ -46,7 +46,7 @@ export class ScheduleProcessController {
             throw new BadRequestException('Wrong process id');
         }
 
-        const scheduleProcessWithUser = { ...scheduleProcess, user: request.user, process};
+        const scheduleProcessWithUser = { ...scheduleProcess, user: request.user, process };
```
star-in-the-jar, line 49: imo change name to the one not including WITH 
& why with 'user' if it also contains 'process'?

PR #228, opened 2023-06-05
[RPA 764] Added Banner on Blog

MRSIMM0:
## Description

image

## Types of changes

---


runbotics/runbotics-orchestrator-ui/src/src-landing/components/BlogHeadImage/BlogHeadImage.module.scss
```diff
@@ -2,45 +2,48 @@
...
     &::before {
-        content: '';
         position: absolute;
+
         width: 100%;
         height: 100%;
+
         z-index: $header-z-index;
-        background: transparent url('/images/blog/mask.svg');
+        background: #FBB040;
```
preludium: isn't it in `_palette.scss`? Looks a bit like our primary colour if its not maybe we should use a primary

runbotics/runbotics-orchestrator-ui/src/src-landing/components/BlogHeadImage/BlogHeadImage.module.scss
```diff
@@ -2,45 +2,48 @@
...
     &::after{
         @include font-face-base;
         position: absolute;
-        z-index: 1;
-        content: "RunNews";
+
         bottom: -40px;
         right: 70px;
 
+        z-index: 1;
```
preludium, line 38: add [isolation: isolate](https://developer.mozilla.org/en-US/docs/Web/CSS/isolation) on parent to prevent unwanted behaviour due to zindex

runbotics/runbotics-orchestrator-ui/src/src-landing/components/BlogHeadImage/BlogHeadImage.tsx
```diff
@@ -1,9 +1,13 @@
 import {FC} from 'react';
 
+import Image from 'next/image';
+
 import styles from './BlogHeadImage.module.scss';
 
 const BlogHeadImage: FC = () => (
-    <div className={styles.image} />
+    <div className={styles.image}>
+        <Image src='/images/banners/blog-header.png' alt='banner' style={{objectFit: 'cover'}} fill={true}/>
```
preludium: - import src like here `runbotics\runbotics-orchestrator-ui\src\src-landing\components\PostHeader\PostHeader.tsx`
- object-fit assign via css class
- if a property is true constantly then there is no need to write it jest leave property name

preludium (changes requested):

preludium (approved):

PR #235, opened 2023-06-14
[RPA-743] Suggest vars created by other actions' output

star-in-the-jar:
## Description

Fixed this problem:
BEFORE

Right now the vars created by some actions as theirs' output are suggested. **Also are limited by scope**:
vars_fix
vars_fix_2

## Types of changes

---


star-in-the-jar (commented):
Self cr

preludium (approved):

PR #262, opened 2023-07-07
Feature/RPA-805 Pomijanie procesów gości

star-in-the-jar:
## Description
Now, processes created by guests are not visible for normal users (are visible only for admins).

The guests' processes are no longer visible:
- on processes list page (the grid), 
- through searchbar,
- in action StartProcess.

## Types of changes

---


runbotics-orchestrator/src/main/java/com/runbotics/domain/UserAuthority.java
```diff
@@ -0,0 +1,26 @@
...
+    private String authorityName;
+
+    public Long getUserId() {
+        return userId;
+    }
+
+    public void setUserId(Long userId) {
+        this.userId = userId;
+    }
+}
```
preludium: the linking tables should not exist as domains. Create Many-to-Many field in the Authority entity and link there users just like we do with feature keys

preludium: then just get all guests authorities and you have all users joined to it

runbotics-orchestrator/src/main/java/com/runbotics/service/ProcessQueryService.java
```diff
@@ -192,4 +200,21 @@ private Specification<Process> isCreatedByUser() {
...
+        List<ProcessDTO> filteredProcesses = new ArrayList<>();
+
+        if(!hasRequesterRoleAdmin) {
+            for (ProcessDTO process : processes) {
+                if(!guestIds.contains(process.getCreatedBy().getId())) {
+                    filteredProcesses.add(process);
+                }
+            }
+            return filteredProcesses;
+        }
```
preludium: all of this you can replace with a single stream chain

preludium (changes requested):

runbotics-orchestrator/src/main/java/com/runbotics/service/ProcessQueryService.java
star-in-the-jar: DONE

star-in-the-jar (commented):
self reviewed

runbotics-orchestrator/src/main/java/com/runbotics/domain/UserAuthority.java
star-in-the-jar: done

preludium (approved):

PR #274, opened 2023-07-14
[RPA-844] Bug fix for bad substitution of objects and lists to # variable

marcin-depcik:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-common/src/Expressions/index.ts
+++ b/runbotics/runbotics-common/src/Expressions/index.ts
@@ -68,10 +68,28 @@ export class Expressions {
             };
 
             response.jexl = true;
-            response.result = templatedString.replace(jexlPattern, (expressionMatch, innerProperty) => {
-                const evaluatedProperty = Jexl.evalSync(innerProperty, jexlContext);
-                return evaluatedProperty;
-            })
+
+            let isPropertyArray = false;
+            const property = templatedString.replace(jexlPattern, (expressionMatch, innerProperty) => {
+                    const evaluatedProperty = Jexl.evalSync(innerProperty, jexlContext);
+                    
+                    isPropertyArray = Array.isArray(evaluatedProperty);
+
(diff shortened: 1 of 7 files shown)


runbotics/runbotics-common/src/Expressions/index.ts
```diff
@@ -68,10 +68,28 @@ export class Expressions {
             };
 
             response.jexl = true;
-            response.result = templatedString.replace(jexlPattern, (expressionMatch, innerProperty) => {
-                const evaluatedProperty = Jexl.evalSync(innerProperty, jexlContext);
-                return evaluatedProperty;
-            })
+
+            let isPropertyArray = false;
```
star-in-the-jar: avoid using let. function programming/useState only

star-in-the-jar: I think the simplest is replacing:
if (isPropertyArray) {

with:
if(Array.isArray(JSON.parse(property)))

runbotics/runbotics-common/src/Expressions/index.ts
```diff
@@ -68,10 +68,28 @@ export class Expressions {
...
+                response.result = JSON.parse(property);
+            } else if (property === "true") {
+                response.result = true;
+            } else if (property === "false") {
+                response.result = false;
+            } else if (property === "null") {
+                response.result = null;
+            } else {
+                response.result = property;
+            }
```
star-in-the-jar: maybe use switch?

star-in-the-jar (changes requested):

preludium (dismissed):

runbotics/runbotics-common/src/Expressions/index.ts
marcin-depcik: In this case I need information about evaluatedProperty specifically from inside of replace function. Without JSON.stringify, replace will return an array parsed to String, which causes the bug.

star-in-the-jar (dismissed):

PR #284, opened 2023-07-17
[RPA-865] Add default null value for required input fields

marcin-depcik:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-orchestrator-ui/src/src-app/Actions/general.actions.ts
+++ b/runbotics/runbotics-orchestrator-ui/src/src-app/Actions/general.actions.ts
@@ -23,7 +23,7 @@ const getGeneralActions: () => Record<string, IBpmnAction> = () => ({
                                 additionalProperties: {
                                     mainFieldLabel: translate('Process.Details.Modeler.Actions.General.ConsoleLog.Message'),
                                     subFieldLabel: translate('Process.Details.Modeler.Actions.General.ConsoleLog.Variable'),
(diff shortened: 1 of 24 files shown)


star-in-the-jar (dismissed):
imo everything's all right if:
- it has been tested,
- not throwing any [type] errors,
- we only want to set null for required fields and nothing was missed.

what's importand it does not affect existing actions, only the new ones

runbotics/runbotics-orchestrator-ui/src/src-app/Actions/general.actions.ts
```diff
@@ -23,7 +23,7 @@ const getGeneralActions: () => Record<string, IBpmnAction> = () => ({
                                 additionalProperties: {
                                     mainFieldLabel: translate('Process.Details.Modeler.Actions.General.ConsoleLog.Message'),
                                     subFieldLabel: translate('Process.Details.Modeler.Actions.General.ConsoleLog.Variable'),
-                                    isRequired: true,
+                                    isRequired: false,
```
star-in-the-jar, line 26: why is it no longer required?

star-in-the-jar (approved):

runbotics/runbotics-orchestrator-ui/src/src-app/Actions/general.actions.ts
star-in-the-jar: does console.log without any message make sense? we always want to display some message, usually variable

PR #290, opened 2023-07-18
[RPA-865-2] Force validation of Target input for browser actions

marcin-depcik:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessBuildView/Modeler/ActionFormPanel/widgets/BrowserTargetWidget/BrowserTargetWidget.tsx
+++ b/runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessBuildView/Modeler/ActionFormPanel/widgets/BrowserTargetWidget/BrowserTargetWidget.tsx
@@ -52,11 +52,14 @@ const BrowserTargetWidget: FC<WidgetProps> = (props) => {
             const prefix = Object.keys(PREFIX_OPTIONS).find((option) => props.value.startsWith(option)) || 'None';
             const value = prefix === 'None' ? props.value : props.value.replace(prefix, ''); 
             setFormState({ prefix: prefix, value: value});
+            onChangeWrapper(prefix, value);
         }
     }, []);
 
     const onChangeWrapper = (prefix: string, value: string) => {
-        if(prefix === 'None') {
+        if (!prefix || !value) {
+            props.onChange(null);
+        } else if (prefix === 'None') {
             props.onChange(value);
         } else{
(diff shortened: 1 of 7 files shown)


runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessBuildView/Modeler/ActionFormPanel/widgets/BrowserTargetWidget/BrowserTargetWidget.tsx
```diff
@@ -52,11 +52,14 @@ const BrowserTargetWidget: FC<WidgetProps> = (props) => {
...
         }
     }, []);
 
     const onChangeWrapper = (prefix: string, value: string) => {
-        if(prefix === 'None') {
+        if (!prefix || !value) {
+            props.onChange(null);
+        } else if (prefix === 'None') {
             props.onChange(value);
         } else{
```
star-in-the-jar: missing space after else

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessBuildView/Modeler/ActionFormPanel/widgets/BrowserTargetWidget/BrowserTargetWidget.tsx
```diff
@@ -52,11 +52,14 @@ const BrowserTargetWidget: FC<WidgetProps> = (props) => {
...
             props.onChange(prefix + value);
```
star-in-the-jar: what if value is "sth" and prefix is false/null? this way -> propsOnChange(null)

marcin-depcik: We want that when either prefix or value is falsy both inputs are invalid, means locked till user will set proper values. Method props.onChange(null) will force this behavior. Moreover, onMounted, prefix always gets value from PREFIX_OPTIONS, so I think scenario that prefix is falsy will never happen.

star-in-the-jar: LeonardoDicaprioSoldGifGIF

star-in-the-jar (dismissed):

star-in-the-jar (approved):

PR #299, opened 2023-07-19
[RPA-884] Add a variable - email of the person starting the process

marcin-depcik:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-common/src/model/api/process-instance.model.ts
+++ b/runbotics/runbotics-common/src/model/api/process-instance.model.ts
@@ -18,7 +18,7 @@ export interface IProcessInstance {
     bot?: IBot;
     error?: string | null;
     trigger?: ITriggerEvent;
-    triggerData?: EmailTriggerData | ITriggerData | unknown;
+    triggerData?: EmailTriggerData | UserTriggerData | unknown;
     warning?: boolean;
 }
 
@@ -29,7 +29,7 @@ export interface EmailTriggerData {
     bccRecipients?: string[];
 }
 
-export interface ITriggerData {
+export interface UserTriggerData {
     userEmail?: string;
 }
 
--- a/runbotics/runbotics-common/src/model/api/process-instance.model.ts
+++ b/runbotics/runbotics-common/src/model/api/process-instance.model.ts
@@ -18,7 +18,7 @@ export interface IProcessInstance {
     bot?: IBot;
     error?: string | null;
     trigger?: ITriggerEvent;
-    triggerData?: EmailTriggerData | unknown;
+    triggerData?: EmailTriggerData | ITriggerData | unknown;
     warning?: boolean;
 }
 
@@ -27,6 +27,10 @@ export interface EmailTriggerData {
     sender: string;
     ccRecipients?: string[];
     bccRecipients?: string[];
+}
+
+export interface ITriggerData {
+    userEmail?: string;
 }
 
 export enum ProcessInstanceStatus {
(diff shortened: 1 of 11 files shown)


runbotics/runbotics-common/src/model/api/process-instance.model.ts
```diff
@@ -29,6 +29,10 @@ export interface EmailTriggerData {
     bccRecipients?: string[];
 }
 
+export interface ITriggerData {
```
star-in-the-jar: prefixing interfaces with I is an antipattern

star-in-the-jar: speaking of naming how is it different from EmailTriggerData? It's good to include that in the name

runbotics/runbotics-desktop/src/core/bpm/runtime/runtime.service.ts
```diff
@@ -382,6 +382,7 @@ export class RuntimeService implements OnApplicationBootstrap, OnModuleDestroy {
             variables: {
                 ...request.variables,
                 tempFolder: this.getTempDirPath(),
+                userEmail: request.triggerData.userEmail,
```
star-in-the-jar: here you use request.triggerData.userEmail but in the interface userEmail is optional

whether add ? here or remove ? from the interface

runbotics/runbotics-scheduler/src/queue/process/process.controller.ts
```diff
@@ -43,6 +43,7 @@ export class ProcessController {
                 input,
                 user: request.user,
                 trigger: { name: TriggerEvent.MANUAL },
+                triggerData: { userEmail: request.user.email }
```
star-in-the-jar, line 46: same as above

star-in-the-jar (changes requested):

runbotics/runbotics-scheduler/src/queue/process/process.controller.ts
marcin-depcik: if you look at the parameter type of createInstantJob function, you will see that field triggerData is optional field, so when I want pass this field I don't have to use "?"

runbotics/runbotics-desktop/src/core/bpm/runtime/runtime.service.ts
```diff
@@ -376,12 +376,14 @@ export class RuntimeService implements OnApplicationBootstrap, OnModuleDestroy {
...
 
         const services = this.createEngineExecutionServices(processInstanceId);
+        const triggerData = request.triggerData;
 
         const engineExecutionOptions: BpmnEngineExecuteOptions = {
             services,
             variables: {
                 ...request.variables,
                 tempFolder: this.getTempDirPath(),
+                ...(triggerData && 'userEmail' in triggerData && { userEmail: triggerData.userEmail }),
```
star-in-the-jar: if it works in case triggerData has no userEmail or even if triggerData does not exists at all than merge

runbotics/runbotics-desktop/src/core/bpm/runtime/runtime.service.ts
```diff
@@ -376,12 +376,14 @@ export class RuntimeService implements OnApplicationBootstrap, OnModuleDestroy {
         };
 
         const services = this.createEngineExecutionServices(processInstanceId);
+        const triggerData = request.triggerData;
```
star-in-the-jar: request?.triggerData -> in the interface you specified that triggerData is optional parameter

star-in-the-jar (approved):

PR #303, opened 2023-07-20
Release 2.7.1

preludium:


runbotics/runbotics-orchestrator-ui/src/src-app/components/guards/GuestGuard.tsx
```diff
@@ -2,23 +2,39 @@ import { FC, VFC, useEffect } from 'react';
...
+        if (!user.roles.includes(Role.ROLE_GUEST)) {
+            router.replace('/app/processes');
+        }
+
+        if (user.roles.includes(Role.ROLE_GUEST) && router.query.guest !== 'true') {
+            dispatch(authActions.logout())
+                .then(() => {
+                    router.replace('/');
+                });
+        }
```
star-in-the-jar, line 36: yesterday i got error "role" does not exist in "user" -> i think it's possible on first guest creation
in this case this goes: router.replace('/app/processes'); and we've got an infinity loader.

but if you've tested it with rushx start:dev and rushx start, keep it.

star-in-the-jar:
everything else looks all right

PR #307, opened 2023-07-26
[RPA-848] Allow easy access to the loop iterator

marcin-depcik:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-orchestrator-ui/src/src-app/hooks/useProcessActionVariables.ts
+++ b/runbotics/runbotics-orchestrator-ui/src/src-app/hooks/useProcessActionVariables.ts
@@ -35,7 +35,7 @@ const useProcessActionVariables = (selectedElementParentId?: string) => {
         elements.forEach(
             element => {
                 if (element.actionId === 'loop.loop') {
-                    localLoops.push(element);
+                    localLoops.push({ ...element, scopeId: element.actionId });
                     return;
                 }
                 localVarsAssigningActions.push({ ...element, scopeId });
@@ -95,7 +95,7 @@ const useProcessActionVariables = (selectedElementParentId?: string) => {
             );
             
         return ({
-            varsAssigningActions: [...localVarsAssigningActions, ...scopeLoopsActions],
+            varsAssigningActions: [...localVarsAssigningActions, ...scopeLoopsActions, ...localLoops],
             updatedScopeTree
         });
     };
@@ -115,7 +115,7 @@ const useProcessActionVariables = (selectedElementParentId?: string) => {
 
         
         if (!allLocalVarsActions || !canvas) {
-            return { inputActionVariables: [], outputActionVariables: [] };
+            return { inputActionVariables: [], outputActionVariables: [], loopVariables: [] };
         }
 
         /**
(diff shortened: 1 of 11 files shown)


runbotics/runbotics-orchestrator-ui/src/src-app/hooks/useProcessActionVariables.ts
```diff
@@ -183,16 +183,21 @@ const useProcessActionVariables = (selectedElementParentId?: string) => {
...
             .flat();
 
-        if(!selectedElementParentId) return { inputActionVariables: inputActionVars, outputActionVariables: outputActionVars };
+        const loopIds = allVarsAssigningActions
+            .filter((element: ScopedModdleElement) => element.actionId === 'loop.loop')
+            .map((element: ScopedModdleElement) => element.id);
+
+        const loopIterator = loopIds.includes(selectedElementParentId) ? [{ name: 'iterator', value: 'variable' }] : [];
+
+        if(!selectedElementParentId) return { inputActionVariables: inputActionVars, outputActionVariables: outputActionVars, loopVariables: [] };
```
star-in-the-jar: I think you can move 

if(!selectedElementParentId) return { inputActionVariables: inputActionVars, outputActionVariables: outputActionVars, loopVariables: [] };

over 

const loopIds = allVarsAssigningActions.......

because if selectedElementParentId does not exist you don't return neither loopIds nor loopIterator so unnecessary calculations take place

runbotics/runbotics-orchestrator-ui/src/src-app/hooks/useProcessActionVariables.ts
```diff
@@ -183,16 +183,21 @@ const useProcessActionVariables = (selectedElementParentId?: string) => {
             .filter((item: ActionVariableObject[]) => item.length > 0)
             .flat();
 
-        if(!selectedElementParentId) return { inputActionVariables: inputActionVars, outputActionVariables: outputActionVars };
+        const loopIds = allVarsAssigningActions
+            .filter((element: ScopedModdleElement) => element.actionId === 'loop.loop')
+            .map((element: ScopedModdleElement) => element.id);
+
+        const loopIterator = loopIds.includes(selectedElementParentId) ? [{ name: 'iterator', value: 'variable' }] : [];
```
star-in-the-jar: does it work if you have loop inside another loop? If yes, won't they be both "#{iterator}"?

it may be hard to differ them - should we suffix them with e.g. scopeId? like #{iterator-s2kfw43e}? I don't think so - still don't know which loop is which.
Maybe we should give access only to the closest scope iterator (for case when loop is inside another loop)? - i suggest to discuss it with Mikołaj.

preludium (dismissed):

runbotics/runbotics-orchestrator-ui/src/src-app/hooks/useProcessActionVariables.ts
marcin-depcik: Each loop recognizes its own iterator, so even in nested loops it works perfectly. I've also updated documentation, how to deal with nested loops using variables.

preludium (approved):

PR #320, opened 2023-08-03
[RPA-902] Add action Clear cells

star-in-the-jar:
## Description

image
image

image
image

yellow color only for demo presentation. not for production

errors handled:
image

## Types of changes

---

### DIFF
--- a/runbotics/runbotics-orchestrator-ui/src/src-app/Actions/excel.actions.ts
+++ b/runbotics/runbotics-orchestrator-ui/src/src-app/Actions/excel.actions.ts
@@ -281,6 +281,39 @@ const getExcelActions: () => Record<string, IBpmnAction> = () => ({
             },
         },
     },
+    'excel.clearCells': {
+        id: 'excel.clearCells',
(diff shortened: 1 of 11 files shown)


star-in-the-jar (commented):
self review

runbotics/runbotics-orchestrator-ui/src/src-app/Actions/excel.actions.ts
```diff
@@ -136,6 +136,39 @@ const getExcelActions: () => Record<string, IBpmnAction> = () => ({
...
+                            }
+                        },
+                        required: ['targetCells'],
+                    },
+                },
+            },
+            uiSchema: {
+                'ui:order': ['input'],
+            },
+            formData: {},
```
marcin-depcik, line 314: If you checked that there won't be any typeErrors then I think it is ok to leave formData as empty object.

star-in-the-jar: why would it be?
I suppose it should be 
            formData: {
                output: {
                    variableName: undefined,
                }
            },
?

preludium (approved):

PR #321, opened 2023-08-03
[RPA-841] Changed endpoint to delete users by id

Aeri4a:
## Types of changes

---

### DIFF
--- a/runbotics-orchestrator/src/main/java/com/runbotics/repository/ProcessRepository.java
+++ b/runbotics-orchestrator/src/main/java/com/runbotics/repository/ProcessRepository.java
@@ -39,13 +39,13 @@ public interface ProcessRepository extends JpaRepository<Process, Long>, JpaSpec
     Page<Process> findByBotCollection(String collectionName, String userLogin, Pageable pageable);
 
     @Query(
-        value = "SELECT COUNT(process.id) FROM process WHERE process.created_by_id = :id",
-        nativeQuery = true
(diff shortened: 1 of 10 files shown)


runbotics-orchestrator/src/main/java/com/runbotics/repository/ProcessRepository.java
```diff
@@ -38,6 +38,12 @@ public interface ProcessRepository extends JpaRepository<Process, Long>, JpaSpec
     )
     Page<Process> findByBotCollection(String collectionName, String userLogin, Pageable pageable);
 
+    @Query(
+        value = "SELECT COUNT(process.id) FROM process WHERE process.created_by_id = :id",
+        nativeQuery = true
+    )
```
star-in-the-jar: try removing nativeQuery. In case we want to use another DB (like MySQL instead of PostgreSQL, which is slightly probable, but not impossible) this will no longer work

star-in-the-jar: if the language syntax is different e.g. in T-SQL there's TOP but in MySQL there's LIMIT

star-in-the-jar (approved):
check the comment

star-in-the-jar (changes requested):
bump version

star-in-the-jar:
please provide screen shots

star-in-the-jar (approved):
noice

PR #324, opened 2023-08-04
[RPA-895] Updating of process update field

marcin-depcik:
## Types of changes

---


star-in-the-jar (approved):
The only problem I see here is when we add new method it's easy to forget about updating timezone too. But current is probably the best solution

PR #339, opened 2023-08-16
[RPA-842-2] Endpoint to edit users

Aeri4a:
## Types of changes

---


star-in-the-jar (dismissed):
I've read your JIRA comment. We should return error when someone tries to change any of forbidden fields like lastModified date, createdDate, etc. Not setting them is not enough

runbotics-orchestrator/src/main/java/com/runbotics/service/UserService.java
```diff
@@ -316,6 +317,14 @@ public Optional<AdminUserDTO> partialUpdate(AdminUserDTO adminUserDTO) {
...
                     }
 
+                    userRepository
+                        .findOneByEmailIgnoreCase(adminUserDTO.getEmail())
+                        .ifPresent(user -> {throw new EmailAlreadyUsedException();});
+
+                    userRepository
+                        .findOneByLogin(adminUserDTO.getLogin())
+                        .ifPresent(user -> {throw new LoginAlreadyUsedException();});
+
```
preludium: it would be more than welcome to have it as a single sql operation

preludium: it would be more than welcome to have it as a single sql operation

preludium (changes requested):

preludium (approved):

PR #345, opened 2023-08-17
[RPA-907] Adds new excel action: insert rows after

bhszewczyk:
## Description

Allows user to insert rows after specified row number.

Inputs:
Start row: number - mandatory
Numbert of rows: number - mandatory
Worksheet: string - optional

Option #1 - for active sheet
Inputs:
image

Starting file:
image

Outcome:
image

Option #2 - for provided worksheet
Inputs:
image

Starting file:
image

Outcome:
image

Error handling:
Start row and Number of rows has to be a whole positive number
image

## Types of changes

---


runbotics/runbotics-actions-windows/src/automation/office/excel.action-handler.ts
```diff
@@ -259,6 +260,27 @@ export default class ExcelActionHandler extends StatefulActionHandler {
...
+
+            targetWorksheet
+                .Range(
+                    targetWorksheet.Rows(startingRow + 1),
+                    targetWorksheet.Rows(startingRow + rowsNumber))
+                .Insert();
+        } catch (e) {
+            throw new Error(ExcelErrorMessage.insertRowsIncorrectInput(e));
+        }
+    }
```
MRSIMM0, line 279: Is this try catch necessary? You could just throw your desired exception in if above.

MRSIMM0 (changes requested):

runbotics/runbotics-actions-windows/src/automation/office/excel.action-handler.ts
bhszewczyk: This is the convention we have in other excel actions as well so to be consistent I did it that way. In case one input is wrong we show error about all the inputs for the action (and correct values suggestions). In my case e.g. method Insert cannot be done if the starting row is a negative number (it throws exception on that method, I would say it is handled by the excel method - could be simplified explanation, but still) but row could be a negative number and this is not what we expect from the user, so I added throw Error there (without the message, cause catch show the all inputs error message).

MRSIMM0 (approved):

runbotics/runbotics-actions-windows/src/automation/office/excel.action-handler.ts
Klau-Sam: also imo here it should be solved without try and catch, just by checking rowNumber and startingRow, an error could be thrown to clearly inform what we expect from user before calling the Excel method. what also speaks for this solution is that these errors thrown strictly from Excel methods are non-descriptive and actually cause more confusion, as you showed us on ss. perhaps the rest of the actions will be refactored in the future in this regard

runbotics/runbotics-orchestrator-ui/src/src-app/translations/pl/process/actions/excel.ts
```diff
@@ -41,6 +41,10 @@ const excelActionsTranslations: typeof englishExcelActionsTranslations = {
     'Process.Details.Modeler.Actions.Excel.InsertColumns.ColumnsAmount': 'Liczba kolumn',
     'Process.Details.Modeler.Actions.Excel.InsertColumns.Column.Info': 'Oznaczenie literowe lub numer kolumny, np "AZ" lub 25.',
     'Process.Details.Modeler.Actions.Excel.InsertColumns.Amount.Info': 'Liczba kolumn do wstawienia.',
+    'Process.Details.Modeler.Actions.Excel.InsertRowsAfter.Label': 'Wstaw wiersze po',
+    'Process.Details.Modeler.Actions.Excel.InsertRows.RowsNumber': 'Liczba rzędów do wstawienia',
+    'Process.Details.Modeler.Actions.Excel.InsertRows.Row.Info': 'Początkowy numer rzędu, np. 5',
```
Klau-Sam: numer początkowego wiersza will maybe sound better? rzędy -> wiersze - what do you think?

runbotics/runbotics-actions-windows/src/automation/office/excel.action-handler.ts
bhszewczyk: Agree! Changed to guards.

PR #346, opened 2023-08-17
[RPA-873] Added Delete worksheet action

MRSIMM0:
## Motivation and Context

image
image

## Types of changes

---


runbotics/runbotics-actions-windows/src/automation/office/excel.action-handler.ts
```diff
@@ -211,72 +186,59 @@ export default class ExcelActionHandler extends StatefulActionHandler {
...
+            if (!Array.isArray(input.targetCells)) targetWorksheet.Range(input.targetCells).Clear();
+            else for (const cellCoordinate of input.targetCells) targetWorksheet.Range(cellCoordinate).Clear();
         } catch (e) {
             throw new Error(ExcelErrorMessage.clearCellsIncorrectInput(e));
         }
     }
 
+    async deleteWorksheet(input: ExcelDeleteWorksheetActionInput): Promise<void> {
+
+        const targetWorksheet = this.session.Worksheets(input?.worksheet ?? this.session.ActiveSheet.Name);
```
bhszewczyk: I think we may want to have the worksheet mandatory in this action - since it is deletion I think we should make user super aware about it and have him point what to delete. We will avoid situation where there were some changes made in previous actions, this worksheet stays active and the next action deletes this worksheet.

MRSIMM0: This was intonational. There is an message that if worksheets name is not given active worksheet is deleted.

bhszewczyk: Yes, there is a message, but I can imagine that when the user adds this action and see no errors before saving he may not read the tooltip. From process creation perspective I would prefer to be forced to provide this info as it will make me fully aware of what I am doing, but if that was intentional and well thought then ok - I rest my case. 😃

runbotics/runbotics-actions-windows/src/automation/office/excel.action-handler.ts
```diff
@@ -47,56 +49,41 @@ export default class ExcelActionHandler extends StatefulActionHandler {
...
 
             for (let rowIdx = startRow; rowIdx <= endRow; rowIdx++) {
                 const rowValues: unknown[] = [];
                 for (let columnIdx = startColumn; columnIdx <= endColumn; columnIdx++) {
-                    rowValues.push(
-                        targetWorksheet
-                            .Cells(rowIdx, columnIdx)
-                            .Value() ?? ''
-                    );
+                    rowValues.push(targetWorksheet.Cells(rowIdx, columnIdx).Value() ?? '');
```
Klau-Sam: I think the formatting should stay the same :(

runbotics/runbotics-actions-windows/src/automation/office/excel.action-handler.ts
```diff
@@ -211,72 +183,66 @@ export default class ExcelActionHandler extends StatefulActionHandler {
...
+            if (!Array.isArray(input.targetCells)) targetWorksheet.Range(input.targetCells).Clear();
+            else for (const cellCoordinate of input.targetCells) targetWorksheet.Range(cellCoordinate).Clear();
         } catch (e) {
             throw new Error(ExcelErrorMessage.clearCellsIncorrectInput(e));
         }
     }
 
+    async deleteWorksheet(input: ExcelDeleteWorksheetActionInput): Promise<void> {
+        try {
+            if (input.worksheet === undefined) {
```
Klau-Sam: this will be caught by the catch clause without throwing the error here, but if this input field is required, the form itself will not allow this field to be left blank. maybe it's better to check if the worksheet we want to delete exists at all and give an appropriate message to the user?

Klau-Sam (changes requested):

runbotics/runbotics-orchestrator-ui/src/src-app/translations/pl/process/actions/excel.ts
```diff
@@ -35,6 +35,8 @@ const excelActionsTranslations: typeof englishExcelActionsTranslations = {
     'Process.Details.Modeler.Actions.Excel.RenameWorksheet.Label': 'Zmień nazwę arkusza',
     'Process.Details.Modeler.Actions.Excel.RenameWorksheet.NewName': 'Nowa nazwa',
     'Process.Details.Modeler.Actions.Excel.RenameWorksheet.Worksheet.Info': 'Pozostawienie pustego pola spowoduje zmianę nazwy aktywnego arkusza',
+    'Process.Details.Modeler.Actions.Excel.DeleteWorksheet.Label': 'Usuń Arkusz',
```
Klau-Sam: Dear Mr. Arkusz :D

runbotics/runbotics-actions-windows/src/automation/office/excel.action-handler.ts
```diff
@@ -249,6 +250,20 @@ export default class ExcelActionHandler extends StatefulActionHandler {
         }
     }
 
+    async deleteWorksheet(input: ExcelDeleteWorksheetActionInput): Promise<void> {
+        if (input.worksheet === undefined) {
```
Klau-Sam: redundant imo

Klau-Sam (approved):

PR #347, opened 2023-08-18
[bugfix/RPA-922] Action's output is unavailable inside a loop

marcin-depcik:
## Description

image

### Before fix
image

### After fix
image

## Types of changes

---


runbotics/runbotics-desktop/src/core/bpm/CamundaExtension.js
```diff
@@ -6,6 +6,8 @@ export default {
     moddleOptions,
 };
 
+const outputVariables = {};
```
preludium: this approach can lead to sharing variables between different process executions. Propose a way to handle it within a single process execution but remember that we can have subprocesses

preludium (changes requested):

preludium (approved):

PR #362, opened 2023-08-28
[RPA-944] Fixes issue where negative number of rows/columns was accepted

bhszewczyk:
Number of columns/rows to insert provided in the input cannot be negative - wrong input error message is thrown.

image

## Types of changes

---

### DIFF
--- a/runbotics/runbotics-actions-windows/src/automation/office/excel.action-handler.ts
+++ b/runbotics/runbotics-actions-windows/src/automation/office/excel.action-handler.ts
@@ -206,7 +206,7 @@ export default class ExcelActionHandler extends StatefulActionHandler {
         try {
             const column = this.getColumnCoordinate(input.column);
             const amount = input.amount;
-            if (!this.isPositiveInteger(amount)) throw new Error();
+            if (this.isNegativeInteger(amount)) throw new Error();
 
             targetWorksheet
                 .Range(targetWorksheet.Columns(column), targetWorksheet.Columns(column + amount - 1))
@@ -223,7 +223,7 @@ export default class ExcelActionHandler extends StatefulActionHandler {
         try {
             const column = this.getColumnCoordinate(input.column);
             const amount = input.amount;
-            if (!this.isPositiveInteger(amount)) throw new Error();
+            if (this.isNegativeInteger(amount)) throw new Error();
 
             targetWorksheet
                 .Range(targetWorksheet.Columns(column + 1), targetWorksheet.Columns(column + amount))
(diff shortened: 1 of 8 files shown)


runbotics/runbotics-actions-windows/src/automation/office/excel.action-handler.ts
```diff
@@ -618,4 +628,8 @@ export default class ExcelActionHandler extends StatefulActionHandler {
         }
         throw new Error(ExcelErrorMessage.tableNotFoundIncorrectInput());
     }
+
+    private isPositiveInteger(number: number): boolean {
```
MRSIMM0: In every occurrence of this function you negate the result. So maybe isNegativeInteager function would be better?

runbotics/runbotics-actions-windows/src/automation/office/excel.action-handler.ts
```diff
@@ -218,10 +219,11 @@ export default class ExcelActionHandler extends StatefulActionHandler {
     async insertColumnsAfter(input: ExcelInsertColumnsActionInput): Promise<void> {
         if (input.worksheet) this.checkIsWorksheetNameCorrect(input.worksheet, true);
         const targetWorksheet = this.session.Worksheets(input?.worksheet ?? this.session.ActiveSheet.Name);
-
+         
         try {
             const column = this.getColumnCoordinate(input.column);
             const amount = input.amount;
+            if (!this.isPositiveInteger(amount)) throw new Error();
```
preludium: provide error message plx

runbotics/runbotics-actions-windows/src/automation/office/excel.action-handler.ts
```diff
@@ -618,4 +628,8 @@ export default class ExcelActionHandler extends StatefulActionHandler {
         }
         throw new Error(ExcelErrorMessage.tableNotFoundIncorrectInput());
     }
+
+    private isPositiveInteger(number: number): boolean {
+        return (number > 0 && Number.isInteger(number))
```
preludium: isInteger check should occur before checking value

preludium (approved):

runbotics/runbotics-actions-windows/src/automation/office/excel.action-handler.ts
bhszewczyk: It is not being provided not to double the error message thrown in a catch. :)

runbotics/runbotics-actions-windows/src/automation/office/excel.action-handler.ts
bhszewczyk: Ok, changed it. The only thing is that 0 is neither negative or positive number so isNegative now check if the number is below 0, so 0 is acceptable (previously it wasn't), action will be executed, but nothing will happen. Checking if is positive allowed me to exclude 0 as well.

PR #398, opened 2023-09-12
[RPA-931] Refactor of Cloud Excel actions

preludium:


runbotics/runbotics-desktop/src/action/action.module.ts
```diff
@@ -41,7 +40,6 @@ const ALL_ACTION_HANDLERS = [
     LoopActionHandler,
     MailActionHandler,
     JavaScriptActionHandler,
-    SharepointExcelActionHandler,
```
preludium, line 44: no need to include it here, already exported via MicrosoftModule

PR #402, opened 2023-09-15
[feature/RPA-989] Create action Sharepoint Excel: Delete columns

marcin-depcik:
## Description

image

## Types of changes

---


runbotics/runbotics-common/src/actions/actions.ts
```diff
@@ -11,7 +11,8 @@ export enum CloudPath {
 export enum ActionRegex {
     VARIABLE_NAME = '^[a-zA-Z0-9_]*$',
     EXCEL_WORKSHEET_NAME = '^[^\\[\\]\\*\\?\\/\\\\\\:\\|]*$',
-    EXCEL_CELL_ADDRESS = '^[a-zA-Z]+\\$?\\d+'
+    EXCEL_CELL_ADDRESS = '^[a-zA-Z]+\\$?\\d+',
```
star-in-the-jar: imo this also should be '^[A-Z]+\\$?\\d+'

runbotics/runbotics-common/src/actions/actions.ts
```diff
@@ -11,7 +11,8 @@ export enum CloudPath {
...
+    EXCEL_COLUMN_NAME = "^[A-Z]$",
```
star-in-the-jar: it's not right. column can be also BAF but here can be only 1-letter

runbotics/runbotics-orchestrator-ui/src/src-app/translations/en/process/process.json
```diff
@@ -326,6 +326,10 @@
     "Process.Details.Modeler.Actions.SharePointExcel.OpenFileFromRoot.FilePath.Info": "Refers to the location path, where the starting point is the folder, where the bot lands after login to the account, e.g. folderName/fileName.xlsx",
     "Process.Details.Modeler.Actions.SharePointExcel.OpenFileFromRoot.ListName.Info": "The default name is 'Documents'",
     "Process.Details.Modeler.Actions.CloudExcel.CloseSession.Label": "Close session",
+    "Process.Details.Modeler.Actions.CloudExcel.DeleteColumns.Label": "Delete columns",
+    "Process.Details.Modeler.Actions.CloudExcel.DeleteColumns.StartColumn": "Start column (value must be [A-Z])",
+    "Process.Details.Modeler.Actions.CloudExcel.DeleteColumns.EndColumn": "End column (value must be [A-Z])",
```
star-in-the-jar: In all excel desktop actions we have example value (and most of the restrictions) in the tooltip. 
+
[A-Z] contains regex expression which is not familiar to non-programmers. imo "uppercase letters A-Z" + example in the tooltip are enough

runbotics/runbotics-orchestrator-ui/src/src-app/translations/en/process/process.json
```diff
@@ -326,6 +326,10 @@
...
+    "Process.Details.Modeler.Actions.CloudExcel.DeleteColumns.EndColumn.Info": "If not specified, only the starting column will be deleted",
```
star-in-the-jar: 👌

runbotics/runbotics-desktop/src/action/microsoft/excel/automation/cloud-excel.action-handler.ts
```diff
@@ -59,6 +59,19 @@ export class CloudExcelActionHandler extends StatefulActionHandler {
         return this.excelService.setRange(this.session, input.range, input.values);
     }
 
+    deleteColumns(input: SharepointTypes.CloudExcelDeleteColumnsActionInput) {
+        const startColumn = input.startColumn.match(ActionRegex.EXCEL_COLUMN_NAME);
+        const endColumn = input.endColumn ? input.endColumn.match(ActionRegex.EXCEL_COLUMN_NAME) : startColumn;
```
star-in-the-jar, line 68: right now this won't work for e.g. range AB-AG

runbotics/runbotics-desktop/src/action/microsoft/excel/excel.service.ts
```diff
@@ -179,6 +179,20 @@ export class ExcelService {
...
+    /**
+     * @see https://learn.microsoft.com/en-us/graph/api/range-delete?view=graph-rest-1.0&tabs=http
+     */
+    deleteColumns(session: ExcelSession, columnRange: string) {
+        return this.microsoftGraphService.
+            post(
+                this.createWorkbookUrl(session, `/worksheets/${session.worksheetName}/range(address='${columnRange}')/delete`),
+                {
+                    ...this.getSessionHeader(session),
+                    shift: 'left'
```
star-in-the-jar, line 202: what is it for?

star-in-the-jar (changes requested):

runbotics/runbotics-common/src/actions/actions.ts
preludium: why dollar sign? and imo there is no difference between lower and upper case

preludium (changes requested):

runbotics/runbotics-common/src/actions/actions.ts
marcin-depcik: I think this pattern is completely wrong. It verifies the coordinates of a cell, e.g. C1, but also CaB123, so imo it should match a string with two substrings. The first part matches letters and the second part matches numbers. My suggestion is a pattern that looks as follows: ^[a-zA-Z]+(\d+)$

runbotics/runbotics-desktop/src/action/microsoft/excel/excel.service.ts
marcin-depcik: This function sends request to delete column/columns in worksheet. You can read more here: https://learn.microsoft.com/en-us/graph/api/range-delete?view=graph-rest-1.0&tabs=http

runbotics/runbotics-common/src/actions/actions.ts
star-in-the-jar: I'd only added '+' after [A-Z].  What @marcin-depcik suggests is best solution but imo without the brackets: ^[a-zA-Z]+\d+$. Unless you want to immediately split letters and numbers which sounds like a good idea, but then imo both should be in brackets ^([a-zA-Z]+)(\d+)$

star-in-the-jar: > why dollar sign? and imo there is no difference between lower and upper case

the dollar sign means end of line

marcin-depcik: I mean the address 'CaB123' should be valid input, correct me if I am wrong.

runbotics/runbotics-orchestrator-ui/src/src-app/translations/pl/process/process.json
```diff
@@ -330,6 +330,11 @@
     "Process.Details.Modeler.Actions.SharePointExcel.OpenFileFromRoot.FilePath.Info": "Odnosi się do ścieżki, gdzie punktem początkowym jest folder, w którym bot ląduje po zalogowaniu się do konta, np. nazwaFolderu/nazwaPliku.xlsx",
     "Process.Details.Modeler.Actions.SharePointExcel.OpenFileFromRoot.ListName.Info": "Domyślna nazwa to 'Documents'",
     "Process.Details.Modeler.Actions.CloudExcel.CloseSession.Label": "Zakończ sesję",
+    "Process.Details.Modeler.Actions.CloudExcel.DeleteColumns.Label": "Usuń kolumny",
+    "Process.Details.Modeler.Actions.CloudExcel.DeleteColumns.StartColumn": "Kolumna początkowa",
+    "Process.Details.Modeler.Actions.CloudExcel.DeleteColumns.EndColumn": "Kolumna końcowa",
+    "Process.Details.Modeler.Actions.CloudExcel.DeleteColumns.StartColumn.Info": "Nazwa kolumny początkowej (np. A)",
+    "Process.Details.Modeler.Actions.CloudExcel.DeleteColumns.EndColumn.Info": "Nazwa kolumny końcowej (np. H), jeśli nie określono wartości, usunięta zostanie tylko kolumna początkowa",
```
star-in-the-jar: desktop excel examples are in form ..., np. sth - no brackets, yes comma
+
grammar suggestion:
Nazwa kolumny końcowej, np. H. Jeżeli nie określono wartości, usunięta zostanie tylko kolumna początkowa

star-in-the-jar (approved):

preludium (approved):

PR #404, opened 2023-09-19
[RPA-956] Guest IP address anonymization

marcin-depcik:
## Description

DONE:
1. Changes in GuestServiceImpl
2. Add new database changelog to change the guest column name to 'ip_hash'

## Types of changes

---

### DIFF
--- a/runbotics-orchestrator/src/main/java/com/runbotics/service/impl/GuestServiceImpl.java
+++ b/runbotics-orchestrator/src/main/java/com/runbotics/service/impl/GuestServiceImpl.java
@@ -39,7 +39,9 @@ public class GuestServiceImpl implements GuestService {
     }
 
     public boolean verifyGuestLimit(String guestIp) {
-        Optional<Guest> guestOptional = guestRepository.findById(guestIp);
+        Optional<Guest> guestOptional = guestRepository.findAll().stream()
+            .filter(guest -> passwordEncoder.matches(guestIp, guest.getIp()))
+            .findFirst();
         return guestOptional.isEmpty();
     }
 
@@ -76,7 +78,7 @@ public class GuestServiceImpl implements GuestService {
 
     private Guest createGuest(String guestIp, User user) {
         var guest = new Guest();
-        guest.setIp(guestIp);
+        guest.setIp(passwordEncoder.encode(guestIp));
         guest.setExecutionsCount(0);
         guest.setUser(user);
         return guest;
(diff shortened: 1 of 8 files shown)


runbotics-orchestrator/src/main/resources/config/liquibase/changelog/20230919606868_changelog.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:pro="http://www.liquibase.org/xml/ns/pro" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/pro http://www.liquibase.org/xml/ns/pro/liquibase-pro-4.1.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+    <changeSet author="depcikm" id="20230919606868-1">
+        <preConditions onFail="MARK_RAN">
+            <and>
+                <tableExists tableName="guest"/>
+                <not>
```
Aeri4a, line 8: Does it mean - do it when table exists AND column NOT exists? Don't we check if column actually exists to make sure that we can change its name?

runbotics-orchestrator/src/main/resources/config/liquibase/changelog/20230919606868_changelog.xml
```diff
@@ -0,0 +1,14 @@
...
+                    <columnExists tableName="guest" columnName="ip_hash"/>
+                </not>
+            </and>
+        </preConditions>
```
preludium: include condition about ip column existance

preludium (approved):

runbotics-orchestrator/src/main/java/com/runbotics/service/impl/GuestServiceImpl.java
```diff
@@ -39,7 +39,9 @@ public GuestServiceImpl(
     }
 
     public boolean verifyGuestLimit(String guestIp) {
-        Optional<Guest> guestOptional = guestRepository.findById(guestIp);
+        Optional<Guest> guestOptional = guestRepository.findAll().stream()
+            .filter(guest -> passwordEncoder.matches(guestIp, guest.getIp()))
+            .findFirst();
```
Aeri4a, line 44: It looks good for me. I'm just wondering if there is another way to find just one guest user or it makes no difference?

marcin-depcik: The reason why I did it this way is that bcrypt generates a unique hash every time, so we are not able to make such a query to the database that will check if the guest with that ip_hash already exists

PR #417, opened 2023-09-26
[RPA-978] Refactor caching mechanism on the page, blog and single post page

marcin-depcik:
## Description

DONE:
1. Remove code responsible for updating posts each day.
2. Refactor caching mechanism no the home, blog and single post page.

## Types of changes

---


runbotics/runbotics-orchestrator-ui/src/pages/_app.tsx
```diff
@@ -40,6 +41,12 @@ moment.locale(DEFAULT_LANG);
 // Client-side cache, shared for the whole session of the user in the browser.
 const clientSideEmotionCache = createEmotionCache();
 
+(async function initializeBlogCache() {
+    if (!isCached('en') || !isCached('pl')) {
+        await recreateCache();
+    }
+})();
```
preludium, line 48: are you sure it runs at app start?

preludium: to me its more like if someone will visit any page of the app

runbotics/runbotics-orchestrator-ui/src/pages/blog/index.tsx
```diff
@@ -45,22 +42,16 @@ const BlogPage: VFC<Props> = ({ posts, categories, tags, page, featuredPost }) =
...
 
-    let cache = getBlogMainCache(locale as Language);
-
-    if (!cache) {
-        cache = await recreateCache(locale as Language);
-
-    } else {
-        res.setHeader('X-Cache', 'HIT');
-    }
-
```
preludium, line 63: why remove header?

runbotics/runbotics-orchestrator-ui/src/pages/blog/post/[slug].tsx
```diff
@@ -27,27 +22,16 @@ interface Params extends Record<string, string> {
...
-    params,
-    locale,
-}) => {
-    let post: BlogPost | undefined;
+export const getServerSideProps: GetServerSideProps<Props, Params> = async ({ params, locale }) => {
+    const language = locale as Language;
 
-    if (isCacheUpToDate((locale ? locale : 'en') as Language)) {
-        post = getSinglePostCache(locale as Language, params.slug);
-        res.setHeader('X-Cache', 'HIT');
```
Comment (preludium) [inline runbotics/runbotics-orchestrator-ui/src/pages/blog/post/[slug].tsx:39]: same here

runbotics/runbotics-orchestrator-ui/src/pages/index.tsx
```diff
@@ -16,14 +16,9 @@ interface Props {
 
 const IndexPage: VFC<Props> = ({ blogPosts }) => <MainView blogPosts={blogPosts} />;
 
-export const getServerSideProps: GetServerSideProps<Props> = async ({ res, locale }) => {
-    let cache = getBlogMainCache(locale as Language);
-
-    if (!cache) {
-        cache = await recreateCache(locale as Language);
-    } else {
-        res.setHeader('X-Cache', 'HIT');
```
preludium, line 25: and here

preludium (changes requested):

runbotics/runbotics-orchestrator-ui/src/pages/_app.tsx
marcin-depcik: I tested this solution in development and production mode. In development mode, the compilation is executed only when the user enters the page, so this logic is run only when the user visits the page. In production mode, the application runs based on already compiled code, so the logic here is executed after the application is launched without the user having to visit the page

runbotics/runbotics-orchestrator-ui/src/pages/blog/index.tsx
marcin-depcik: image

Above code inside _app.tsx is executed each time user entered the page, so to remove code duplication, I decided to move it.

runbotics/runbotics-orchestrator-ui/src/pages/_app.tsx
marcin-depcik: I can provide an additional condition to clarify that this logic is intended for `production` purposes only, eg:
`if (process.env.NODE_ENV === 'production' && (!isCached('en') || !isCached('pl')))`

runbotics/runbotics-orchestrator-ui/src/pages/blog/index.tsx
preludium: wrong, we want it to be only on blog related pages

preludium (changes requested):

preludium (approved):

PR #429, opened 2023-10-16
[RPA-1044-2] Cloud File move file action - remove output

Aeri4a:
…e action

## Types of changes

---


preludium (changes requested):
You should remove return statement from action handler. Service methods should be reusable. We just dont want to return any value from move file actions handler

preludium (approved):

PR #450, opened 2023-10-26
[RPA-1058] Add bot disconnect email to owner/admins

marcin-depcik:
## Types of changes

---


runbotics/runbotics-scheduler/src/nest-mailer/nest-mailer.module.ts
```diff
@@ -0,0 +1,32 @@
...
+                    auth: {
+                        user: process.env.MAIL_USERNAME,
+                        pass: process.env.MAIL_PASSWORD,
+                    },
+                },
+                defaults: {
+                    from: process.env.MAIL_USERNAME,
+                },
+            }),
+        }),
```
preludium: use ServerConfigService

preludium: does it work even if we don't provide those envs? or what is happening in this scenario?

runbotics/runbotics-scheduler/src/nest-mailer/nest-mailer.module.ts
```diff
@@ -0,0 +1,32 @@
...
+        UserModule,
+        Logger,
+    ],
+    providers: [NestMailerService],
+    exports: [NestMailerService],
+})
+export class NestMailerModule {}
```
preludium: we just use MailerModule from nest but its our implementation then lets create custom name for it. I propose to rename existing MailModule and its service to `MailTriggerModule` and `MailTriggerService` because its has only 1 purpose. This Module/Service will be reusable so lets reserve for it names like MailModule/MailService

runbotics/runbotics-scheduler/src/nest-mailer/nest-mailer.service.ts
```diff
@@ -0,0 +1,62 @@
...
+    private readonly logger = new Logger(NestMailerService.name);
+
+    constructor(
+        private readonly mailerService: MailerService,
+    ) {}
+
+    public async sendMail(input: SendMailInput) {
+        const sendMailOptions: ISendMailOptions = {
+            to: input.to,
+            from: process.env.MAIL_USERNAME, // sender address
```
preludium: obvious comment

preludium (changes requested):

runbotics/runbotics-scheduler/src/nest-mailer/nest-mailer.module.ts
marcin-depcik: Can I use ServerConfigService here while dynamically initializing MailerModule with forRootAsync method? One idea I have right now is to make NestMailerModule as dynamic module and then inject ServerConfigService there and use its methods. What do you think about this approach?

marcin-depcik: About env variables, I tested it and all errors are handled by `.catch()` while executing ` this.mailerService.sendMail()`. All errors, including those caused by invalid env variables will be caught there safely and then displayed as error message.

preludium: here is what you need:
image


> Can I use ServerConfigService here while dynamically initializing MailerModule with forRootAsync method? One idea I have right now is to make NestMailerModule as dynamic module and then inject ServerConfigService there and use its methods. What do you think about this approach?

preludium (approved):

preludium (approved):

PR #458, opened 2023-11-08
[RPA-1113] Speed up process deletion

damlit:
Speed up process deletion

## Description

Entities related to process should be removed in background (on different thread).

## Types of changes

---


runbotics-orchestrator/src/main/resources/config/liquibase/changelog/20231108538060_changelog.xml
```diff
@@ -0,0 +1,65 @@
+<databaseChangeLog xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+
+    <changeSet author="litwind" id="20231108538060-1">
```
damlit, line 5: 1st and 2nd changeSets are to other bug (RPA-1112).

runbotics-orchestrator/src/main/java/com/runbotics/service/impl/ProcessServiceImpl.java
```diff
@@ -274,6 +273,13 @@ public void delete(Long id) {
         tagService.deleteUnusedTags(remainingTags);
     }
 
+    @Async
+    public void deleteProcessLeftovers() {
+        log.debug("Deleting leftovers old Processes has started");
+        processInstanceRepository.deleteInstancesWithoutProcess();
+        log.debug("Deleting leftovers old Processes has finished");
```
preludium: to me, its more readable if we start logs with result status

preludium (approved):

PR #469, opened 2023-11-21
[RPA-629] Exclusive Gateway improvements

damlit:
Changed exclusiveGateway action to flow after it & fixed in UI

## Description

These four improvements has been implemented -> https://all41.atlassian.net/browse/RPA-629?focusedCommentId=28835 .

## Types of changes

---


runbotics/runbotics-desktop/src/core/bpm/runtime/runtime.service.ts
```diff
@@ -184,7 +188,7 @@ export class RuntimeService implements OnApplicationBootstrap, OnModuleDestroy {
             }
         );
 
-        listener.on('activity.start', (api: BpmnExecutionEventMessageApi) => {
+        listener.on('activity.start', (api: BpmnExecutionEventMessageExtendedApi) => {
```
preludium, line 191: is there any event that uses not extended version?

runbotics/runbotics-desktop/src/core/bpm/runtime/runtime.service.ts
```diff
@@ -554,4 +575,16 @@ export class RuntimeService implements OnApplicationBootstrap, OnModuleDestroy {
...
+        const gatewayName = api.name ? api.name : api.id;
+        this.idNameCacheService.setValue(api.id, gatewayName);
+    };
+
+    private isSequenceFlowAfterGateway = (api: BpmnExecutionEventMessageExtendedApi) => (
+        api.content &&
+        api.content.type === BpmnElementType.SEQUENCE_FLOW &&
+        api.content.isSequenceFlow &&
+        api.content.sourceId.includes('Gateway_')
+    );
```
preludium: app here only `api.content` if there is interface for it already

runbotics/runbotics-desktop/src/core/websocket/bpmn/runtime-subscriptions.service.ts
```diff
@@ -16,14 +16,16 @@ import dayjs from 'dayjs';
...
 
 @Injectable()
 export class RuntimeSubscriptionsService {
     constructor(
         @Inject(forwardRef(() => WebsocketService))
         private readonly websocketService: WebsocketService,
         private readonly runtimeService: RuntimeService,
-        private readonly loopHandlerService: LoopHandlerService
+        private readonly loopHandlerService: LoopHandlerService,
+        private readonly idNameCacheService:IdNameCacheService
```
preludium: space after colon

runbotics/runbotics-desktop/src/core/websocket/cache/cache.module.ts
```diff
@@ -0,0 +1,12 @@
...
+
+
+@Global()
+@Module({
+    providers: [
+        IdNameCacheService,
+    ],
+    exports: [IdNameCacheService],
+})
+export class CacheModule {}
```
preludium: imo `cache` can be misleading, its more like storage or smthing like it

runbotics/runbotics-desktop/src/core/websocket/cache/id-name-cache.service.ts
```diff
@@ -0,0 +1,24 @@
+import { Injectable } from '@nestjs/common';
+
+@Injectable()
+export class IdNameCacheService {
```
preludium: this class is not specific in any way so imo it's name should stay general like `StorageService` which in fact is already in the config module but I like your implementation much more.

preludium: imo the best way to do it its like you have a single implementation of generic storage service and when you need something specific then you extend it with proper typings. Then you can name it more specific

runbotics/runbotics-desktop/src/core/websocket/cache/cache.module.ts
preludium: do not make this module global imo

runbotics/runbotics-desktop/src/core/websocket/websocket.module.ts
```diff
@@ -36,6 +37,7 @@ import { RuntimeSubscriptionsService } from './bpmn/runtime-subscriptions.servic
             },
         }),
         forwardRef(() => CoreModule),
+        CacheModule,
```
preludium: if its global then no need to import it afaik

preludium (changes requested):

preludium (approved):

PR #470, opened 2023-11-22
[feature/RPA-1105] Add action to export excel data to html table

marcin-depcik:
## Description

image

## Types of changes

---


runbotics/runbotics-orchestrator-ui/src/src-app/Actions/ExcelActions/index.ts
```diff
@@ -45,7 +46,8 @@ const excelActions = {
     getActionWorksheetExists,
     getActionDeleteRows,
     getActionRunMacro,
-    getActionExportToCsv
+    getActionExportToCsv,
+    getRangeToHtmlAction
```
preludium: naming here is different than for the rest of actions

runbotics/runbotics-common/src/actions/actions.ts
```diff
@@ -204,7 +205,8 @@ export enum ExcelAction {
     DELETE_ROWS = 'excel.deleteRows',
     SAVE = 'excel.save',
     CLOSE = 'excel.close',
-    EXPORT_TO_CSV = 'excel.exportToCsv'
+    EXPORT_TO_CSV = 'excel.exportToCsv',
+    RANGE_TO_HTML_TABLE = 'excel.rangeToHtmlTable'
```
preludium: Imo it is better to call it "Export HTML table". The `range` is not important at the beginning, the user will find out what is required to be passed to the action when he/she goes to the properties

preludium (changes requested):

preludium (approved):

PR #471, opened 2023-11-23
[RPA-620] fixed to long variables

damlit:
Bug with unexpected WS bot has disconnected.

## Description

The reason of bug was to long payload -> 2 steps:
1. it doesn't send outputs for loops (they had all iterations outputs joint in one big output (in my case it was over 2mln bytes).
2. increased WS server limit (1MB to 2MB)

## Types of changes

---

### DIFF
--- a/runbotics/runbotics-desktop/src/core/websocket/bpmn/runtime-subscriptions.service.ts
+++ b/runbotics/runbotics-desktop/src/core/websocket/bpmn/runtime-subscriptions.service.ts
@@ -43,9 +43,7 @@ export class RuntimeSubscriptionsService {
                         created: dayjs().toISOString(),
                         processInstance: event.processInstance,
                         executionId: event.activity.executionId,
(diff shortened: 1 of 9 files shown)


runbotics/runbotics-desktop/src/core/websocket/bpmn/runtime-subscriptions.service.ts
```diff
@@ -239,4 +238,14 @@ export class RuntimeSubscriptionsService {
             this.websocketService.emitMessage(message);
         });
     }
+
+    private sanitizeVariable(variableToCheck: string): string {
+        if (variableToCheck && variableToCheck.length > 10000)
+            return JSON.stringify({ message: 'Exceeded max length' });
+        return variableToCheck;
+    }
```
preludium: imo it should accept `any` and do the stringify here

preludium (approved):

PR #489, opened 2023-12-15
[RPA-1167] Fixed handling variables sent in attendance form

Aeri4a:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-scheduler/src/queue/queue.service.ts
+++ b/runbotics/runbotics-scheduler/src/queue/queue.service.ts
@@ -182,14 +182,15 @@ export class QueueService implements OnModuleInit {
     private async handleAttendedProcess(process: IProcess, input?: ProcessInput) {
         if (!process.isAttended) return;
 
-        if (!input?.variables || typeof process.executionInfo !== 'string') {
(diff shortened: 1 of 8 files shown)


runbotics/runbotics-scheduler/src/queue/queue.service.ts
```diff
@@ -182,14 +182,14 @@ export class QueueService implements OnModuleInit {
     private async handleAttendedProcess(process: IProcess, input?: ProcessInput) {
         if (!process.isAttended) return;
 
-        if (!input?.variables || Object.keys(input.variables).length === 0) {
-            const err = 'You haven\'t provided variables for attended process';
+        if (!input?.variables || typeof process.executionInfo !== 'string') {
+            const err = 'You haven\'t provided variables or proper form for attended process';
             this.logger.error(`Failed to add new instant job for process "${process.name}". ${err}`);
             throw new BadRequestException(err);
         }
```
preludium: it will throw error when we dont include even empty object `variables` in the request body. When someone won't include variables then there will be no body

preludium (changes requested):

preludium (approved):

PR #564, opened 2024-02-14
feat(runbotics-desktop): [RPA-1314] EnvServices refactor

damlit:
## Description
Example is in comment section of https://all41.atlassian.net/browse/RPA-1313

## Motivation and Context

Simplifier use of services. It's possible to use services by # and variables without any prefixes.

## Types of changes

---


runbotics/runbotics-desktop/src/core/bpm/CustomServices.ts
```diff
@@ -36,118 +33,43 @@ export const customServices: Record<string, Function> = {
...
     },
-    push: (arg1: any, arg2: any) => {
-        if (!arg1) {
-            return [arg2];
+    push: (array: any[], newElement: any) => {
+        if (!array) {
+            return [newElement];
         }
-        return arg1.concat([arg2]);
+        return array.concat([newElement]);
```
preludium, line 55: here you cant be sure that an array is a type of array, check it plx

runbotics/runbotics-desktop/src/core/bpm/CustomServices.ts
```diff
@@ -36,118 +33,43 @@ export const customServices: Record<string, Function> = {
...
-    comaSeparatedToArray: (s: any) => {
-        return s.split(',');
+    splitAndPick: (arg1: string, splitter: string, pick: number) => {
+        return arg1.split(splitter)[pick];
     },
-    withDefault: (def: any, val: any) => {
-        //console.log("with default:", def, val);
-        return val !== undefined && val !== null && val !== '' ? val : def;
+    parseJson: (json: any) => {
+        return JSON.parse(json);
```
preludium: parameter must be a string otherwise should throw an error

runbotics/runbotics-desktop/src/core/bpm/CustomServices.ts
```diff
@@ -36,118 +33,43 @@ export const customServices: Record<string, Function> = {
...
     },
-    getCurrentDate: (format: any) => {
+    getCurrentDate: () => {
```
preludium: maybe let's have an optional parameter `format`

runbotics/runbotics-desktop/src/core/bpm/runtime/runtime.service.ts
```diff
@@ -588,6 +588,17 @@ export class RuntimeService implements OnApplicationBootstrap, OnModuleDestroy {
...
+                        const processInstance = this.processInstances[processInstanceId];
+                        this.processEventBus.publish({
+                            processInstanceId: processInstanceId,
+                            eventType: ProcessInstanceStatus.ERRORED,
+                            processInstance: {
+                                ...processInstance,
+                                status: ProcessInstanceStatus.ERRORED,
+                                error: `Inputs in next step: ${(e as Error)?.message}`,
+                            },
+                        });
```
preludium: we also publish error in the listener, so it will be done twice. If this code is mandatory for catching variables/services resolve issues then you should create some kind of if guard to send only one message

preludium (approved):

PR #571, opened 2024-02-22
[RPA-1308] schedule processes with attended form v1

damlit:
## Description

Schedule process for processes with initial values has to be possible.

## Motivation and Context

TODO

## Types of changes

---


runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessRunView/ManageProcessForm.tsx
```diff
@@ -9,18 +9,24 @@ import If from '#src-app/components/utils/If';
...
     const { enqueueSnackbar } = useSnackbar();
     const { process } = useSelector((state) => state.process.draft);
-    const isScheduled = process?.schedules?.length > 0;
+    const { schedules } = useSelector(scheduleProcessSelector);
     const [modalOpen, setModalOpen] = useState(false);
     const openModal = () => setModalOpen(true);
     const closeModal = () => setModalOpen(false);
 
+    // eslint-disable-next-line require-await
+    const removeSchedulesWithOldInputVariables = async () => schedules
```
preludium: old? more like if there is no input variables then remove schedules
imo you cant be sure that you will have schedules on the UI because it is only fetched when user visits run tab.
meaning, it won't do anything in the worst-case scenario. There should be an endpoint on backend to handle it

preludium: tbh I thought we will disable those stale schedules not delete them

runbotics/runbotics-scheduler/src/queue/processor/scheduler.processor.ts
```diff
@@ -172,4 +172,8 @@ export class SchedulerProcessor {
             );
         }
     }
+
+    private isIncorrectJobForProcessType(job: Job, process: IProcess) {
+        return (Boolean(job.data.input.variables) && !process.isAttended) || (!job.data.input.variables && process.isAttended);
+    }
```
preludium: it won't stop stale form schedules. Use validation we have for process start endpoints

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessRunView/ScheduleProcess.tsx
```diff
@@ -58,34 +61,52 @@ const ScheduleProcess: FC<ScheduleProcessProps> = ({ onProcessScheduler }) => {
...
     };
 
+    const handleRunAttendedProcess = (variables: Record<string, any>) => {
+        reset();
+        closeModal();
+        onProcessScheduler({
+            cron,
+            variables: JSON.stringify(variables),
+        });
+    };
```
preludium: imo you dont need another state for this. get it from useForm

preludium (changes requested):

runbotics/runbotics-scheduler/src/queue/processor/scheduler.processor.ts
```diff
@@ -102,30 +98,20 @@ export class SchedulerProcessor {
...
-
-        this.logger.log(
-            `[Q Process] Process "${process.name}" freed the queue | JobID: `,
-            job.id
-        );
-
-        return processInstanceIdentifier;
+        if (!isScheduledProcess(job.data)) {
+            return await this.runProcess(process, job);
+        } else {
```
preludium: you don't need `else` if you return in the if first clause

runbotics/runbotics-scheduler/src/queue/processor/scheduler.processor.ts
```diff
@@ -102,30 +98,20 @@ export class SchedulerProcessor {
...
+            this.queueService.handleAttendedProcess(process, job.data.input)
+                .then(async () => {
+                    return await this.runProcess(process, job);
```
preludium: no need to await on return

runbotics/runbotics-scheduler/src/queue/processor/scheduler.processor.ts
```diff
@@ -102,30 +98,20 @@ export class SchedulerProcessor {
...
+                })
+                .catch((err) => {
+                    this.logger.warn(
+                        `[Q Process] Process "${process.name}" has scheduled with/without input variables for other type of process. ${err.message}`
+                    );
+                    return null;
```
preludium: you must throw an error here or return rejected promise otherwise you won't trigger right flow

runbotics/runbotics-scheduler/src/queue/processor/scheduler.processor.ts
preludium, line 1: verify that every path works as intended (1) no bots (2) timeout and so on

preludium: it is an important piece of code here, ideal it would be to create tests for it

runbotics/runbotics-scheduler/src/queue/queue.service.ts
```diff
@@ -179,8 +179,14 @@ export class QueueService implements OnModuleInit {
...
-    private async handleAttendedProcess(process: IProcess, input?: ProcessInput) {
-        if (!process.isAttended) return;
+    async handleAttendedProcess(process: IProcess, input?: ProcessInput) {
+        if (!process.isAttended && !input.variables) return;
+
+        if (!process.isAttended && input.variables) {
+            const err = 'You have provided variables for non attended process';
+            this.logger.error(`Failed to add new instant job for process "${process.name}". ${err}`);
+            throw new BadRequestException(err);
+        }
```
preludium: imo it should not be an error. the process does not force you to provide variables if its not attended but having those should not throw an error - let it pass

preludium (changes requested):

preludium (approved):

PR #593, opened 2024-03-19
[RPA-1247] Fix unreadable truncated action output

star-in-the-jar:
## Description
1. As before if output is smaller than 100_000 chars it's not modified
2. if output is bigger than 100_000 chars every array is being truncated to 2 elements and returned as object
3. if output after truncating arrays is bigger than 100_000 then it's being cut as it was before and returned as object (unfortunately)

for every truncated array/json there's appropriate information "[truncated]"

## Types of changes

---


star-in-the-jar (commented):
Self cr

marcin-depcik (approved):

PR #594, opened 2024-03-22
[RPA-1260.3] Show only sharable users' emails when editing collection

star-in-the-jar:
## Types of changes

---

### DIFF
--- a/runbotics-orchestrator/src/main/java/com/runbotics/service/UserService.java
+++ b/runbotics-orchestrator/src/main/java/com/runbotics/service/UserService.java
@@ -375,6 +375,12 @@ public class UserService {
     }
 
     @Transactional(readOnly = true)
+    public List<User> getNonAdminUsers() {
+    List<User> activeUsersWithIds = userRepository.findAllActivatedNonAdmins();
+        return activeUsersWithIds.stream().filter(user -> !hasAdminRole(user)).collect(Collectors.toList());
+    }
+
+    @Transactional(readOnly = true)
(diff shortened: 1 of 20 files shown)


star-in-the-jar (commented):
Self cr

runbotics-orchestrator/src/main/java/com/runbotics/service/UserService.java
```diff
@@ -374,6 +374,12 @@ public Page<AdminUserDTO> getAllManagedUsers(Pageable pageable) {
         return userRepository.findAll(pageable).map(AdminUserDTO::new);
     }
 
+    @Transactional(readOnly = true)
+    public List<User> getNonAdminUsers() {
+        List<User> activeUsersWithIds = userRepository.findAllActivatedNonAdmins();
+        return activeUsersWithIds.stream().filter(user -> !hasAdminRole(user)).collect(Collectors.toList());
```
marcin-depcik: Do we need to filter, even if you query for non-administrator users?

marcin-depcik (approved):

runbotics-orchestrator/src/main/java/com/runbotics/service/UserService.java
preludium: if you need to filter your query result it means its a bad query

star-in-the-jar: yeah it's redundant

PR #619, opened 2024-04-23
[RPA-1342-3] Process access based on process collections

Aeri4a:
## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

---

### DIFF
--- a/runbotics-orchestrator/src/main/java/com/runbotics/service/impl/ProcessServiceImpl.java
+++ b/runbotics-orchestrator/src/main/java/com/runbotics/service/impl/ProcessServiceImpl.java
@@ -69,16 +69,17 @@ public class ProcessServiceImpl implements ProcessService {
     public ProcessDTO save(ProcessDTO processDTO) {
         log.debug("Request to save Process : {}", processDTO);
         User user = userService.getUserWithAuthorities().get();
+        Long processId = processDTO.getId();
 
         Process process = processMapper.toEntity(processDTO);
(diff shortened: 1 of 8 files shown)


runbotics-orchestrator/src/main/java/com/runbotics/service/impl/ProcessServiceImpl.java
```diff
@@ -59,18 +62,26 @@ public ProcessServiceImpl(
...
         process.setUpdated(ZonedDateTime.now());
         if (process.getId() == null) {
-            User user = userService.getUserWithAuthorities().get();
             process.setCreated(ZonedDateTime.now());
             process.setCreatedBy(user);
             process.setEditor(user);
+        } else {
+            Process existingProcess = processRepository.findById(
+                processDTO.getId()).orElseThrow(() -> new BadRequestAlertException("Cannot find process with this id", ENTITY_NAME, "processNotFound")
+            );
```
marcin-depcik: for code readability maybe it is better to assign it to `processId` variable?

marcin-depcik (approved):
bump version

runbotics-orchestrator/src/main/java/com/runbotics/service/impl/ProcessServiceImpl.java
Aeri4a: You're right. I will do it :)

PR #622, opened 2024-04-24
[RPA-1330] Display nested subprocesses in History table

seul88:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-desktop/src/action/general/general.action-handler.ts
+++ b/runbotics/runbotics-desktop/src/action/general/general.action-handler.ts
@@ -60,12 +60,12 @@ export default class GeneralActionHandler extends StatelessActionHandler {
             }
                 
             const processInstancesIds = Object.keys(this.runtimeService.processInstances);
-            const parentProcessId = processInstancesIds[processInstancesIds.length - 1];
+            const parentProcessId = processInstancesIds[processInstancesIds.length - 1] ?? request.rootProcessInstanceId;
 
             const processInstanceId = await this.runtimeService.startProcessInstance({
                 process: process,
                 variables: request.input.variables,
-                parentProcessInstanceId: parentProcessId ?? request.rootProcessInstanceId,
+                parentProcessInstanceId: parentProcessId,
                 userId: request.userId,
                 orchestratorProcessInstanceId: null,
                 rootProcessInstanceId: request.rootProcessInstanceId ?? request.processInstanceId,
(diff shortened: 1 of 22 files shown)


runbotics/runbotics-desktop/src/action/general/general.action-handler.ts
```diff
@@ -58,16 +58,21 @@ export default class GeneralActionHandler extends StatelessActionHandler {
             if (processSystem !== BotSystem.ANY && processSystem !== system) {
                 reject(new Error(`Process with system (${processSystem}) cannot be run by the bot with system (${system})`));
             }
+                
+            const processInstancesIds = Object.keys(this.runtimeService.processInstances);
+            const parentProcessId = processInstancesIds[processInstancesIds.length - 1];
```
star-in-the-jar: what if processInstancesIds is empty?

star-in-the-jar: ah, I see it lower, but imo let's do it (parentProcessId ?? request.rootProcessInstanceId,) here

runbotics/runbotics-orchestrator-ui/src/src-app/components/tables/Table/Table.tsx
```diff
@@ -48,9 +48,31 @@ const Table = <T extends object>({
...
+            return tableData.map(item => replaceKeyRecursive(item, newKey));
+        }
+      
+        const newObj = {};
+      
+        for (const key in tableData) {
+            if (Object.hasOwnProperty.call(tableData, key)) {
+                const newKeyString = key === subRowProperty ? newKey : key;
+                newObj[newKeyString] = replaceKeyRecursive(tableData[key] as T[], newKey);
+            }
```
star-in-the-jar, line 66: what does it do? can we make it more readable for future devs?

star-in-the-jar (approved):

runbotics/runbotics-orchestrator-ui/src/src-app/components/tables/Table/Table.tsx
seul88: So the recursive function ProcessInstance.utils.ts file creates the tree structure where nested objects are stored as 'subprocesses' arrays. In this recursive function we want to bind those subprocesses to nested rows of table. They are organized in correct structure already, but we need to rename its' keys from value passed as subRowProperty props (in that case -'subprocesses') to 'subRows'.

PR #668, opened 2024-07-22
[RPA-1585-2] Tenants scheduler patch

Aeri4a:
## Types of changes

---

### DIFF
--- a/runbotics-orchestrator/src/main/java/com/runbotics/service/UserService.java
+++ b/runbotics-orchestrator/src/main/java/com/runbotics/service/UserService.java
@@ -335,6 +335,11 @@ public class UserService {
     public Optional<AdminUserDTO> partialUpdate(AdminUserDTO adminUserDTO) {
         log.debug("Request to partially update User : {}", adminUserDTO);
 
+        User requester = getUserWithAuthorities().get();
+        if(!adminUserDTO.isActivated() && Objects.equals(requester.getId(), adminUserDTO.getId())) {
+            throw new BadRequestAlertException("User cannot deactivate itself", ENTITY_NAME, "SelfDeactivate");
+        }
+
         excludeAdminUserDTOFields(adminUserDTO);
         return userRepository
             .findById(adminUserDTO.getId())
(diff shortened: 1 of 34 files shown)


runbotics/runbotics-orchestrator-ui/src/src-app/utils/axios.ts
```diff
@@ -11,4 +12,16 @@ axiosInstance.interceptors.response.use(
...
+
+axiosApi.interceptors.response.use((res) => res, (err) => {
+    if (err.request.status === 401 && !err.request.responseURL.includes('/api/authenticate')) {
+        setAccessToken(null);
+        delete axiosApi.defaults.headers.common.Authorization;
+        window.location.reload();
+    }
+
+    return Promise.reject(err);
+});
```
marcin-depcik, line 25: cool 😎

marcin-depcik (approved):
bump version

runbotics-orchestrator/src/main/java/com/runbotics/service/UserService.java
```diff
@@ -391,6 +390,10 @@ public Optional<AdminUserDTO> partialUpdate(AdminUserDTO adminUserDTO) {
     public Optional<AdminUserDTO> partialUpdateInTenant(AdminUserDTO adminUserDTO) {
         User requester = getUserWithAuthorities().get();
 
+        if(!adminUserDTO.isActivated() && Objects.equals(requester.getId(), adminUserDTO.getId())) {
+            throw new BadRequestAlertException("User cannot deactivate itself", ENTITY_NAME, "SelfDeactivate");
+        }
```
bhszewczyk, line 395: Maybe you could extract this check to a separate method since it is used in partialUpdate and partialUpdateInTenant?

runbotics/runbotics-orchestrator-ui/src/src-app/components/guards/AuthGuard.tsx
```diff
@@ -36,6 +37,7 @@ export const withAuthGuard = ({
     const isAuthenticated = isInitialized && isBrowser && isAuthed;
 
     if (!isAuthenticated) {
+        setAccessToken(null);
```
bhszewczyk, line 40: 👍

bhszewczyk (approved):

runbotics-orchestrator/src/main/java/com/runbotics/service/UserService.java
Aeri4a: It is a good point overall - however I won't do it here because we are migrating from Java and I think I wouldn't add more methods.

PR #672, opened 2024-07-25
[fix/RPA-1614] Upload file not working properly

marcin-depcik:
## Types of changes

---


runbotics/runbotics-desktop/src/action/microsoft/automation/file/cloud-file.action-handler.ts
```diff
@@ -210,23 +210,23 @@ export class CloudFileActionHandler extends StatelessActionHandler {
         };
     }
 
-    private async readLocalFile(path: string) {
-        const content = readFileSync(path);
-        const { mime } = await fromFile(path);
+    private async readLocalFile(filePath: string) {
+        const content = readFileSync(filePath);
+        const mime = mimeTypes.lookup(filePath);
```
Aeri4a: Is function still async?

Aeri4a (approved):

runbotics/runbotics-scheduler/src/queue/process/process-file.service.ts
```diff
@@ -49,17 +49,31 @@ export class ProcessFileService {
...
+            if (!extension) {
+                throw new Error('Unable to determine the extension');
+            }
+
+            return {
+                contentType,
+                extension,
+                fileName,
+            };
         }
```
bhszewczyk: what do you think about mergin check not to repeat the same in 2 ifs?

if (!filenamePrefix.test(base64File) && !standardPrefix.test(base64File)) throw Error...

const fileName = filenamePrefix.test(base64File) ? prefix[0].split(':')[1] randomUUID()
const contentType = prefix[1].split(':')[1];
const extension = mime.extension(contentType);

if (!extension) {
    throw new Error('Unable to determine the extension');
}

return {
    contentType,
    extension,
    fileName,
}

bhszewczyk (changes requested):

runbotics/runbotics-scheduler/src/queue/process/process-file.service.ts
marcin-depcik: In case of `filenamePrefix` check content type is extracted differently `const contentType = prefix[1].split(':')[1];`, but in case of `standardPrefix` it looks like this `const contentType = prefix[0].split(':')[1];`. Small difference but I think it matters.

bhszewczyk: hm, ok. That's fine
I would consider one more line between fileName/contentType then
const contentType = filenamePrefix.test(base64File) ? prefix[1].split(':')[1] : prefix[0].split(':')[1];

bhszewczyk (approved):

PR #677, opened 2024-08-02
[fix/RPA-1576] Services: adding to a variable field removes everything but the variable

marcin-depcik:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-common/src/Expressions/index.ts
+++ b/runbotics/runbotics-common/src/Expressions/index.ts
@@ -10,7 +10,7 @@ const expressionPattern = /\${(.+?)}/;
 const jexlPattern = /#{(.+?)}/;
 const jexlServicePattern = /#{(.+?)\((.*)\)}/;
 const jexlServiceMethodPattern = /(.+?)\((.*)\)/;
-const jexlRecursivePattern = /#\{((?:[^{}]|#\{(?:[^{}]|#\{[^{}]*\})*\})*)\}/;
+const jexlMatchingBracketsPattern = /#\{((?:[^{}]|#\{(?:[^{}]|#\{[^{}]*\})*\})*)\}/;
 const digitPattern = /^-?\d+(\.?\d+)?$/;
 
 export class Expressions {
@@ -71,8 +71,8 @@ export class Expressions {
     }
 
     private static getFullServiceMethodCall(templatedString: string, context, expressionFnContext) {
-        const jexlRecursiveExp = new RegExp(jexlRecursivePattern, 'g');
-        const matches = templatedString.match(jexlRecursiveExp);
+        const jexlMatchingBracketsRegExp = new RegExp(jexlMatchingBracketsPattern, 'g');
+        const matches = templatedString.match(jexlMatchingBracketsRegExp);
         if (!matches) {
             return templatedString;
         }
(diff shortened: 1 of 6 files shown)


runbotics/runbotics-common/src/Expressions/index.ts
```diff
@@ -7,8 +7,10 @@ function _interopRequireDefault(obj) {
 
 const isExpressionPattern = /^\${(.+?)}$/;
 const expressionPattern = /\${(.+?)}/;
-const jexlPattern = /#{(.+?)}/g;
+const jexlPattern = /#{(.+?)}/;
 const jexlServicePattern = /#{(.+?)\((.*)\)}/;
+const jexlServiceMethodPattern = /(.+?)\((.*)\)/;
+const jexlRecursivePattern = /#\{((?:[^{}]|#\{(?:[^{}]|#\{[^{}]*\})*\})*)\}/;
```
star-in-the-jar: is it for variable used inside a variable, like #{ #{myVar1} + #{myVar2}} ?

star-in-the-jar: how many levels does it reach? or is it purely recursive (so infinite levels of recursion depth)?

star-in-the-jar: i it does reach only 5 levels it should have been called e.g. jexlRecursivePatternUpTo5DepthLevel for example. Or there should be a disclaimer somewhere that variables are being supported up to 5-th level of depth. We cannot just hope to not go beyond that border.

star-in-the-jar: Also is jexlRecursivePattern strictly related to jexlPattern? Like the same, but recursive? That's what the names are suggesting. If they are not, the name's backbone (jexl...Pattern) should be different

runbotics/runbotics-common/src/Expressions/index.ts
```diff
@@ -68,21 +70,61 @@ export class Expressions {
...
-            if (jexlServicePattern.test(trimmedArg)) {
-                return Expressions.resolveExpression(trimmedArg, context, expressionFnContext);
-            } else if (Expressions.isVariableArgument(trimmedArg)) {
-                return 'environment.variables.' + trimmedArg;
+    private static getFullServiceMethodCall(templatedString: string, context, expressionFnContext) {
+        const jexlRecursiveExp = new RegExp(jexlRecursivePattern, 'g');
+        const matches = templatedString.match(jexlRecursiveExp);
+        if (!matches) {
+            return templatedString;
+        }
```
star-in-the-jar, line 80: are you sure matches is not an array in any case?

star-in-the-jar: if so then, the !matches.length should have been checked

runbotics/runbotics-common/src/Expressions/index.ts
```diff
@@ -68,21 +70,61 @@ export class Expressions {
...
+                jexlRecursiveExp.exec(templatedString);
+            if (!recursivelyExtractedExpression) continue;
+
+            const extractedMethod =
+                jexlServiceMethodPattern.exec(recursivelyExtractedExpression[1]);
+            if (!extractedMethod) continue;
+
+            const methodName = extractedMethod[1];
+            const args = extractedMethod[2].length
+                ? extractedMethod[2].trim().split(",")
```
star-in-the-jar: create variable for extractedMethod[2] - the code will be more readable
can be done also with:
const [methodName, extractedMethodSecond] = extractedMethod;

star-in-the-jar (approved):

runbotics/runbotics-common/src/Expressions/index.ts
marcin-depcik: `match` retruns results array or `null`

runbotics/runbotics-common/src/Expressions/index.ts
```diff
@@ -7,8 +7,10 @@ function _interopRequireDefault(obj) {
 
 const isExpressionPattern = /^\${(.+?)}$/;
 const expressionPattern = /\${(.+?)}/;
-const jexlPattern = /#{(.+?)}/g;
+const jexlPattern = /#{(.+?)}/;
 const jexlServicePattern = /#{(.+?)\((.*)\)}/;
+const jexlServiceMethodPattern = /(.+?)\((.*)\)/;
+const jexlMatchingBracketsPattern = /#\{((?:[^{}]|#\{(?:[^{}]|#\{[^{}]*\})*\})*)\}/;
```
Aeri4a, line 13: I think I will recommend to use ([^()]*) to exclude brackets. Perhaps it could be useful in the future.

Aeri4a (approved):

PR #688, opened 2024-08-16
[RPA-1670] Perform FeatureKey and Authority migration

star-in-the-jar:
## Types of changes

---


star-in-the-jar (commented):
self review

runbotics/runbotics-scheduler/src/scheduler-database/authority-feature-key/authority-feature-key.entity.ts
```diff
@@ -0,0 +1,22 @@
...
+import { FeatureKey } from '../feature-key/feature-key.entity';
+import { Role as RoleEnum } from 'runbotics-common';
+import { Authority } from '../authority/authority.entity';
+import { FeatureKey as FeatureKeyEnum } from 'runbotics-common';
+
+@Entity({ name: 'authority_feature_key' })
+export class AuthorityFeatureKey {
+    @ManyToOne(() => Authority, { eager: true })
+    @JoinColumn({ name: 'authority', referencedColumnName: 'name' })
+    authorities: Authority[];
```
Aeri4a, line 11: Should we add here cascade? 'onDelete: 'CASCADE' What do you think?

runbotics/runbotics-scheduler/src/database/database.module.ts
```diff
@@ -11,10 +10,11 @@ import { ProcessInstanceEventModule } from './process-instance-event/process-ins
 import { ScheduleProcessModule } from './schedule-process/schedule-process.module';
 import { BotCollectionModule } from './bot-collection/bot-collection.module';
 import { BotSystemModule } from './bot-system/bot-system.module';
-import { FeatureKeyModule } from './feature-key/feature-key.module';
 import { TriggerEventModule } from './trigger-event/trigger-event.module';
 import { ProcessInstanceLoopEventModule } from './process-instance-loop-event/process-instance-loop-event.module';
 import { GuestModule } from './guest/guest.module';
+import { AuthorityModule } from '#/scheduler-database/authority/authority.module';
+import { FeatureKeyModule } from '#/scheduler-database/feature-key/feature-key.module';
```
Aeri4a: I think you could move those entities into scheduler-database module. I know that is not in the same schema, but I think it will be clear if we store directly new migrations in that file. The old one then could be refactored/deleted. This is a small thing, we could discuss it if u want.

Aeri4a (changes requested):
I got two additional thoughts:
1. If it will be possible then we could add enum as a type of FK and Authority name field (but I realize that it could drop column in migration - you could try if you want)
2. Is is good that there is no new migration for Authority & FeatureKey entity, just for their N:N relation?

runbotics/runbotics-scheduler/src/database/database.module.ts
star-in-the-jar: Alright, done

runbotics/runbotics-scheduler/src/scheduler-database/authority-feature-key/authority-feature-key.entity.ts
star-in-the-jar: Yeah, only authority+feature-key pair makes sense in this table, on the other hand we delete authorities hardly ever and there' no endpoint for modification of authorites. So I wouldn't add onDelete cascade, because there's more benefits from it being security feature (in case someone removes authority by mistake, thinking it's unsed) than maintenance errand (when we decide to really remove some authority, which happens hardly ever and is just one additional query)

star-in-the-jar:
> If it will be possible then we could add enum as a type of FK and Authority name field (but I realize that it could drop column in migration - you could try if you want)

I did it in the beginning, but I haven't seen any difference whether it is specified both as entity property - someProperty: SomeEnum - and in config { 'enum': SomeEnum } or only in entityty property.

Intellisense works for both and there's no difference in migrations - there's no enum defined in the migration - column type stays string. I think this may be an TypeORM option for other DBs.

tests:
For example FeatureKey relation with defined "enum" field:
image
Authority without "enum":
image

result:
there'se no difference in migration files

BUT after discussion with @Aeri4a decided to add enum field - there's a chance TypeORM uses it at some point of modifying the db

star-in-the-jar (commented):
self review

star-in-the-jar (commented):
self reviewed migration

Aeri4a (approved):

marcin-depcik (approved):

star-in-the-jar (commented):
self review

PR #698, opened 2024-08-29
[RPA-1462] Credentials in process configure

Aeri4a:
## Description
image
image

image

## Types of changes

---


runbotics/runbotics-common/src/model/api/process-credential.model.ts
```diff
@@ -0,0 +1,19 @@
...
+        template: {
+            id: string;
+            name: string;
+        },
+        collection: {
+            id: string;
+            name: string;
+        }
+    }
+}
```
marcin-depcik: maybe it is better to create separate types for credential and collection?

runbotics/runbotics-orchestrator-ui/src/src-app/Actions/beeOffice.actions.ts
```diff
@@ -1,13 +1,15 @@
 import { BeeOfficeAction, ActionRegex } from 'runbotics-common';
 
+import { ActionCredentialType } from '#src-app/credentials/actionCredentialType.enum';
 import { translate } from '#src-app/hooks/useTranslations';
 
 import { IBpmnAction, Runner } from './types';
 
 // eslint-disable-next-line max-lines-per-function
-const getBeeOfficeActions: () => Record<string, IBpmnAction> = () => ({
+const getBeeOfficeActions: () => Record<string, & Required<Pick<IBpmnAction, 'credentialType'>>> = () => ({
```
marcin-depcik: should be:
`const getBeeOfficeActions: () => Record<string, IBpmnAction & Required<Pick<IBpmnAction, 'credentialType'>>> = () =>({})`

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.tsx
```diff
@@ -0,0 +1,100 @@
...
+    handleCredentialChange: (credential: Credential | undefined) => void;
+}
+
+const CredentialSelectSet: FunctionComponent<CredentialSelectSetProps> = ({
+    credentials, handleCredentialChange
+}) => {
+    const { translate } = useTranslations();
+
+    const [author, setAuthor] = useState<string>(null);
+    const credentialsSortedByAuthor = useMemo(
```
marcin-depcik: `SortedBy` or `FilteredBy`

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.tsx
```diff
@@ -0,0 +1,100 @@
...
+        () => credentials.filter(credential => credential.createdBy.login === author),
+        [author, credentials]);
+    const [collectionName, setCollectionName] = useState<string>(null);
+    const credentialsSortedByCollection = useMemo(
```
marcin-depcik: same as above

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.tsx
```diff
@@ -0,0 +1,100 @@
...
+        () => credentialsSortedByAuthor.filter(credential => credential.collection.name === collectionName),
+        [collectionName, credentialsSortedByAuthor]);
+    const [credentialName, setCredentialName] = useState<string>(null);
+    const credentialSortedByName = useMemo(
```
marcin-depcik: another one

marcin-depcik: imo all those `useMemo` return filtered results

runbotics/runbotics-orchestrator-ui/src/src-app/credentials/getRequiredCredentialsTypesInProcess.ts
```diff
@@ -0,0 +1,15 @@
...
+import actions from '#src-app/Actions';
+import { ActionCredentialType } from '#src-app/credentials/actionCredentialType.enum';
+
+const ACTION_ID_REGEX = /camunda:actionId="([a-zA-Z.]+)"/g;
+
+export const getRequiredCredentialsTypesInProcess = (definition: IProcess['definition']): ActionCredentialType[] =>
+    [
+        ...Array.from(definition.matchAll(ACTION_ID_REGEX), m => m[1])
+            .reduce((set, actionId) =>
+                actions[actionId].credentialType
```
marcin-depcik: I think if `actionId` does not exist in the `actions` then trying to get to `credentialType` will throw an TypeError

runbotics/runbotics-orchestrator-ui/src/src-app/credentials/useRequiredCredentialTypes.ts
```diff
@@ -0,0 +1,13 @@
...
+import { getRequiredCredentialsTypesInProcess } from '#src-app/credentials/getRequiredCredentialsTypesInProcess';
+import { useSelector } from '#src-app/store';
+import { processSelector } from '#src-app/store/slices/Process';
+
+export const useRequiredCredentialTypes = () => {
+    const { draft: { process }} = useSelector(processSelector);
+
+    return useMemo(() =>
+        process.definition ? getRequiredCredentialsTypesInProcess(process.definition) : null,
+    [process]);
```
marcin-depcik: for readability maybe it's better to assign it to named variable?

runbotics/runbotics-orchestrator-ui/src/src-app/providers/ModelerProvider.tsx
```diff
@@ -10,7 +10,7 @@ export const ModelerContext = React.createContext<ModelerContext>(null);
 
 interface ModelerProviderProps {
     modeler?: BpmnModelerType;
-}
+}
```
marcin-depcik: remove whitespace

runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Credentials/Credentials.state.ts
```diff
@@ -0,0 +1,21 @@
...
+    description: string;
+    createdAt: string;
+    updatedAt: string;
+    templateId: string;
+    createdBy: UserDTO;
+    collection: {
+        id: string;
+        name: string;
+    }
+}
```
marcin-depcik: I would use interface created in runbotics-common

runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Process/Process.thunks.ts
```diff
@@ -170,3 +171,13 @@ export const unsubscribeProcessNotifications = ApiTenantResource
 
 export const getProcessSubscriptionInfo = ApiTenantResource
     .get<NotificationProcess[]>('processes/getProcessSubscriptionInfo', `${PROCESS_NOTIFICATION_PATH}/processes`);
+
+export const getProcessCredentials = ApiTenantResource
+    .get<ProcessCredential[]>('processes/getProcessCredentials', `${PROCESS_CREDENTIALS_PATH}/processes`);
+
+export const createProcessCredential = ApiTenantResource
+    .post<void, { credentialId: string; templateName: string; processId: string; }>
```
marcin-depcik: for more than one property I'd create new type

runbotics/runbotics-orchestrator-ui/src/src-app/translations/en/process/configure/credentials.ts
```diff
@@ -0,0 +1,15 @@
+const credentialsTranslations = {
+    'Process.Configure.Credentials.Section.Title': 'Credentials',
+    'Process.Configure.Credentials.Action.PrimaryCredential': 'Primary',
+    'Process.Configure.Credentials.Action.Credential': 'Primary',
```
marcin-depcik: what is the difference?

runbotics/runbotics-orchestrator-ui/src/src-app/translations/pl/process/configure/credentials.ts
```diff
@@ -0,0 +1,17 @@
+import englishCredentialsTranslations from '#src-app/translations/en/process/configure/credentials';
+
+const credentialsTranslations: typeof englishCredentialsTranslations = {
+    'Process.Configure.Credentials.Section.Title': 'Poświadczenia',
+    'Process.Configure.Credentials.Action.PrimaryCredential': 'Główny',
+    'Process.Configure.Credentials.Action.Credential': 'Główny',
```
marcin-depcik: same

runbotics/runbotics-scheduler/src/database/process/process.service.ts
```diff
@@ -30,4 +31,29 @@ export class ProcessService {
...
     }
+
+    async checkAccessAndGetById(processId: number, user: UserEntity) {
+        // TODO: to be updated after process migration to scheduler
+        // Explanation: Process privileges should be based on process collections
+        // and checking privileges should be adjusted
+        return user.authorities.find(authority => authority.name === Role.ROLE_TENANT_ADMIN)
+            ? this.processRepository.findOneByOrFail({ id: processId, tenantId: user.tenantId })
+                .catch(() => {
+                    throw new NotFoundException();
```
marcin-depcik, line 44: throwing error doesn't make sense for me, if no process is found `null` should be returned

runbotics/runbotics-scheduler/src/database/process/process.service.ts
```diff
@@ -30,4 +31,29 @@ export class ProcessService {
...
+                    tenantId: user.tenantId,
+                    isPublic: true
+                },
+                {
+                    id: processId,
+                    tenantId: user.tenantId,
+                    createdBy: { id: user.id }
+                }
+            ]).catch(() => {
+                throw new NotFoundException();
```
marcin-depcik, line 62: same here

runbotics/runbotics-scheduler/src/scheduler-database/credential-collection/credential-collection.service.ts
```diff
@@ -114,6 +114,25 @@ export class CredentialCollectionService {
...
+                    userId: user.id,
+                    privilegeType: PrivilegeType.WRITE,
+                },
+            },
+            relations: [...relations, 'credentials.createdBy', 'credentials.collection'],
+        });
+
+        if (collections.length === 0) {
+            throw new NotFoundException('Could not find any credential collections');
+        }
```
marcin-depcik: just return empty array

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.controller.ts
```diff
@@ -56,8 +57,17 @@ export class CredentialController {
...
+    @User() user: IUser
+  ) {
     this.logger.log('Getting all credentials');
-    return this.credentialService.findAllAccessible(tenantId, user);
+
+    return isCredentialFilterQuery(query)
+      ? this.credentialService.findAllAccessibleByTemplateAndProcess(
+        query.templateName, query.processId, user
+      )
+      : this.credentialService.findAllAccessible(tenantId, user);
```
marcin-depcik: return isCredentialFilterQuery(query)
      ? this.credentialService.findAllAccessibleByTemplateAndProcess(
        query.templateName,
        query.processId,
        user
      )
      : this.credentialService.findAllAccessible(tenantId, user);

I think this is more readable

runbotics/runbotics-scheduler/src/scheduler-database/credential-collection/credential-collection.service.ts
marcin-depcik: is this branch up to date with `develop`? I think you are missing few changes in this service

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
```diff
@@ -74,13 +75,24 @@ export class CredentialService {
...
+        ...credential,
+        createdBy: {
+          id: credential.createdBy.id,
+          login: credential.createdBy.login,
+        },
+        collection: {
+          id: credential.collection.id,
+          name: credential.collection.name
+        }
+      }));
```
marcin-depcik: I guess, you can just use `flatMap()`

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
```diff
@@ -89,6 +101,42 @@ export class CredentialService {
...
 
+  async findAllAccessibleByTemplateAndProcess(templateName: string, processId: string, user: IUser) {
+    const accessibleCollections = await this.collectionService.findAllAccessibleForCredentials(user);
+
+    const credentials = await this.credentialRepo
+      .createQueryBuilder('credentials')
+      .innerJoinAndSelect('credentials.createdBy', 'user')
+      .innerJoinAndSelect(
+        'credentials.collection', 'collection',
+        'collection.id IN (:...accessible)', { accessible: accessibleCollections.map(col=>col.id) }
```
marcin-depcik: I would created additional variable `const accessible`, and remember to add missing spaces in callback

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
```diff
@@ -89,6 +101,42 @@ export class CredentialService {
...
+      ).innerJoinAndSelect(
+        'credentials.template', 'template',
+        'template.name = :templateName', { templateName }
```
marcin-depcik: try to move function params to separate lines

marcin-depcik: apply everywhere, please

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
```diff
@@ -89,6 +101,42 @@ export class CredentialService {
...
+        },
+        collection: {
+          id: credential.collection.id,
+          name: credential.collection.name,
+        },
+      })));
+
+    if (!credentials.length) {
+      throw new NotFoundException('Could not find any credentials in any of the accessible collections');
+    }
```
marcin-depcik: throwing error is not necessary in this case

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
```diff
@@ -89,6 +101,42 @@ export class CredentialService {
...
+        ...credential,
+        createdBy: {
+          id: credential.createdBy.id,
+          login: credential.createdBy.login,
+        },
+        collection: {
+          id: credential.collection.id,
+          name: credential.collection.name,
+        },
+      })));
```
marcin-depcik: please, lint this part

runbotics/runbotics-scheduler/src/scheduler-database/process-credential/process-credential.service.ts
```diff
@@ -0,0 +1,129 @@
...
+        });
+
+        const lastCredentialOrder: number = await this.processCredentialRepository
+            .createQueryBuilder('pc')
+            .select('MAX(pc.order)', 'max')
+            .innerJoin('pc.credential', 'credential')
+            .innerJoin(
+                'credential.template', 'template',
+                'template.name = :templateName', { templateName: processCredentialDto.templateName }
+            ).where('pc.processId = :processId', { processId: processCredentialDto.processId })
```
marcin-depcik: please, lint this part as above

runbotics/runbotics-scheduler/src/utils/typeChecks.ts
```diff
@@ -0,0 +1,12 @@
...
+    templateName: string;
+    processId: string;
+}
+
+export const isCredentialFilterQuery = (
+    query: unknown
+): query is CredentialFilterQuery => query
+&& typeof query === 'object'
+&& 'templateName' in query && typeof query.templateName === 'string'
+&& 'processId' in query && typeof query.processId === 'string';
```
marcin-depcik: maybe move this util to `scheduler-database/credential-collection/utils`?

marcin-depcik (changes requested):

runbotics/runbotics-orchestrator-ui/src/src-app/Actions/asana.actions.ts
```diff
@@ -1,14 +1,14 @@
...
 
+import { ActionCredentialType } from '#src-app/credentials/actionCredentialType.enum';
 import { translate } from '#src-app/hooks/useTranslations';
 
 import { IBpmnAction, Runner } from './types';
 
-
-
-const getAsanaActions: () => Record<string, IBpmnAction> = () => ({
+const getAsanaActions: () => Record<string, IBpmnAction & Required<Pick<IBpmnAction, 'credentialType'>>> = () => ({
```
star-in-the-jar: so now every method getXxxxActions is going to have this type. It'd be nice to have it as a type so if any change occurs it's easy to change in the whole project

runbotics/runbotics-orchestrator-ui/src/src-app/Actions/cloudFile/createCloudFolder.action.ts
```diff
@@ -100,4 +102,4 @@ export const createCloudFolderAction: IBpmnAction = {
             },
         },
     },
-};
+} satisfies IBpmnAction;
```
star-in-the-jar, line 104: nice

runbotics/runbotics-orchestrator-ui/src/src-app/Actions/googleSheets.actions.ts
```diff
@@ -1,14 +1,14 @@
...
 
+import { ActionCredentialType } from '#src-app/credentials/actionCredentialType.enum';
 import { translate } from '#src-app/hooks/useTranslations';
 
 import { IBpmnAction, Runner } from './types';
 
-
-
-const getGoogleSheetsActions: () => Record<string, IBpmnAction> = () => ({
+const getGoogleSheetsActions: () => Record<string, IBpmnAction & Required<Pick<IBpmnAction, 'credentialType'>>> = () => ({
```
star-in-the-jar: universal type here

runbotics/runbotics-orchestrator-ui/src/src-app/Actions/jira-cloud.actions.ts
```diff
@@ -1,12 +1,11 @@
...
 
+import { ActionCredentialType } from '#src-app/credentials/actionCredentialType.enum';
 import { translate } from '#src-app/hooks/useTranslations';
 
 import { IBpmnAction, Runner } from './types';
 
-
-
-const getJiraCloudActions: () => Record<string, IBpmnAction> = () => {
+const getJiraCloudActions: () => Record<string, IBpmnAction & Required<Pick<IBpmnAction, 'credentialType'>>> = () => {
```
star-in-the-jar: universal type here

runbotics/runbotics-orchestrator-ui/src/src-app/Actions/jira-server.actions.ts
```diff
@@ -1,12 +1,11 @@
...
 
+import { ActionCredentialType } from '#src-app/credentials/actionCredentialType.enum';
 import { translate } from '#src-app/hooks/useTranslations';
 
 import { IBpmnAction, Runner } from './types';
 
-
-
-const getJiraServerActions: () => Record<string, IBpmnAction> = () => {
+const getJiraServerActions: () => Record<string, IBpmnAction & Pick<Required<IBpmnAction>, 'credentialType'>> = () => {
```
star-in-the-jar: universal type here

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.styles.ts
```diff
@@ -0,0 +1,14 @@
...
+export const Box = styled.div(({
+    display: 'flex',
+    flexDirection: 'column',
+    alignItems: 'left',
+    width: '70%',
+    gap: '5px',
+}));
+
+export const StyledSelect = styled(Select)(({
+    marginBottom: '20px'
```
star-in-the-jar: if select is inside Box it's better to use padding-bottom in there because of margin overlapping

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.tsx
```diff
@@ -0,0 +1,100 @@
...
+import _ from 'lodash';
+
+import useTranslations from '#src-app/hooks/useTranslations';
+import { Credential } from '#src-app/store/slices/Credentials';
+
+import { Box, StyledSelect } from './CredentialSelectSet.styles';
+
+interface CredentialSelectSetProps {
+    credentials: Credential[];
+    handleCredentialChange: (credential: Credential | undefined) => void;
```
star-in-the-jar: is undefined because of JSONSchemaForm and default value?

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.tsx
```diff
@@ -0,0 +1,100 @@
...
+            <Typography>
+                {translate('Component.CredentialSelectSet.CollectionAuthor')}
+            </Typography>
+            <StyledSelect
+                size='small'
+                value={author}
+                onChange={(e: ChangeEvent<HTMLInputElement>) => { setAuthor(e.target.value); }}
+                disabled={!credentials.length}
+            >
+                {_.uniqBy(credentials, credential => credential.createdBy.login)
```
star-in-the-jar: If I got it right you're showing here only one credential per one author. Why so?

runbotics/runbotics-orchestrator-ui/src/src-app/credentials/getRequiredCredentialsTypesInProcess.ts
star-in-the-jar: Doesn't it mean that the process is outdated (e.g. the action specified there is no longer supported or the id has changed) or invalid? If so the process couldn't be handled by bot, so imo the error can be thrown here, but it's good to name it or to handle it in a way that does not break the whole app)

runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Credentials/Credentials.slice.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const initialState: CredentialsState = {
+    all: [],
+    loading: false
+};
+
+const slice = createSlice({
+    name: 'credentials',
+    initialState,
+    reducers: {},
```
star-in-the-jar, line 17: is it mandatory field?

runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Credentials/Credentials.state.ts
star-in-the-jar: + isn't description optional?

star-in-the-jar: if it's different than one in runbotics-common this should be reflected in the name of the interface

runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Process/Process.thunks.ts
star-in-the-jar: CreateProcessCredentialDto

runbotics/runbotics-orchestrator-ui/src/src-app/translations/en/process/configure/credentials.ts
```diff
@@ -0,0 +1,15 @@
...
+    'Process.Configure.Credentials.Modal.Delete.Title': 'Delete process credential',
+    'Process.Configure.Credentials.Modal.Delete.Content': 'Are you sure you want to delete process credential?',
+    'Process.Configure.Credentials.Modal.Delete.Info.Success': 'Successfully deleted process credential',
+    'Process.Configure.Credentials.Modal.Delete.Info.Error': 'Error occurred while deleting process credential',
```
star-in-the-jar: If we don't know what happened I suggest adding "Unexpected" [error occured ...]

runbotics/runbotics-orchestrator-ui/src/src-app/translations/en/process/configure/credentials.ts
```diff
@@ -0,0 +1,15 @@
...
+    'Process.Configure.Credentials.Modal.Add.Title': 'Add new credential for process action',
+    'Process.Configure.Credentials.Modal.Add.Button.Add': 'Add',
+    'Process.Configure.Credentials.Modal.Add.Info.Success': 'Successfully added process credential',
+    'Process.Configure.Credentials.Modal.Add.Info.Error': 'Error occurred while adding process credential',
```
star-in-the-jar: as above

runbotics/runbotics-orchestrator-ui/src/src-app/utils/types/page.ts
```diff
@@ -22,6 +22,7 @@ export interface PageRequestParams<T extends object = any> {
         equals?: ObjectOf<T>;
         in?: IdentifiersObjectOf<T>;
     };
+    [key: string]: unknown;
```
star-in-the-jar, line 25: what's that needed for?

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentials.utils.ts
```diff
@@ -0,0 +1,52 @@
...
+    }[]
+}> {};
+
+export const sortByActionCredentialType = (
+    credentials: ProcessCredential[],
+    credentialTypes: ActionCredentialType[]
+): ActionCredentials => {
+    const actions: ActionCredentials = credentialTypes.reduce((acc, type) => (acc[type] = [], acc), {});
+    credentials.forEach(pc => {
+        actions[pc.credential.template.name as ActionCredentialType].push({
```
star-in-the-jar: can be this assertion avoided?

runbotics/runbotics-scheduler/src/migrations/1724064431315-ProcessCredentialEntity.ts
```diff
@@ -0,0 +1,40 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class ProcessCredentialEntity1724064431315 implements MigrationInterface {
+    name = 'ProcessCredentialEntity1724064431315'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "scheduler"."process_credential" DROP CONSTRAINT "FK_ad32ad8114ed1f6603ff6fb7c9d"`);
+        await queryRunner.query(`ALTER TABLE "scheduler"."process_credential" DROP CONSTRAINT "FK_588ebc19945e3b661e15dd1fece"`);
+        await queryRunner.query(`DROP INDEX "scheduler"."IDX_588ebc19945e3b661e15dd1fec"`);
+        await queryRunner.query(`DROP INDEX "scheduler"."IDX_ad32ad8114ed1f6603ff6fb7c9"`);
```
star-in-the-jar, line 10: are these added by us or auto-generated?

star-in-the-jar (changes requested):

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.tsx
```diff
@@ -0,0 +1,100 @@
...
+    const { translate } = useTranslations();
+
+    const [author, setAuthor] = useState<string>(null);
+    const credentialsSortedByAuthor = useMemo(
+        () => credentials.filter(credential => credential.createdBy.login === author),
+        [author, credentials]);
+    const [collectionName, setCollectionName] = useState<string>(null);
+    const credentialsSortedByCollection = useMemo(
+        () => credentialsSortedByAuthor.filter(credential => credential.collection.name === collectionName),
+        [collectionName, credentialsSortedByAuthor]);
```
socjalis: some newlines would be nice

socjalis: it's a simple operation, I don't think useMemo is necessary here, as even for hundreds of thousands of credentials the operation time would be negligible. If ever the performance was the issue we could use the for loop instead of `array.filter` to shorten the time for the first render too.

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentialsAddDialog.tsx
```diff
@@ -0,0 +1,99 @@
...
+}
+
+export const ProcessCredentialsAddDialog: FunctionComponent<ProcessCredentialsAddDialogProps> = ({
+    isOpen, handleClose, templateName
+}) => {
+    const { enqueueSnackbar } = useSnackbar();
+    const { id: processId } = useRouter().query;
+    const dispatch = useDispatch();
+    const { all } = useSelector(credentialsSelector);
+    const [pickedCredential, setPickedCredential] = useState<Credential | undefined>();
```
socjalis: Using null for initializing variables that are meant to hold some value later on is more a standard way than undefined

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentialsAddDialog.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+    const handleCredentialChange = (credential: Credential | undefined) => {
+        setPickedCredential(credential);
```
socjalis, line 40: does it need to be a separate function?

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentialsAddDialog.tsx
```diff
@@ -0,0 +1,99 @@
...
+    };
+
+    const handleCredentialAdd = () => {
```
socjalis: handleAddCredential

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentialsAddDialog.tsx
```diff
@@ -0,0 +1,99 @@
...
+        const payload = {
+            processId: String(processId),
+            credentialId: pickedCredential.id,
+            templateName
+        };
+
+        dispatch(processActions.createProcessCredential({ payload }))
```
socjalis, line 50: do you want to close the dialog even when the save fails? would be a good idea to await for this dispatch or move the action logic to async thunk, so someone reading the code would know if it's intentional

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentials.tsx
```diff
@@ -0,0 +1,150 @@
...
+    const actionCredentials = useMemo(
+        () => credentialTypes ? sortByActionCredentialType(processCredentials, credentialTypes) : null,
+        [processCredentials, credentialTypes]
+    );
+
+    const [columns, setColumns] = useState<ActionSortedColumns>([]);
+    const [isAddDialogOpen, setIsAddDialogOpen] = useState(false);
+    const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false);
+    const [currentCredentialId, setCurrentCredentialId] = useState<string>(null);
+    const [currentActionName, setCurrentActionName] = useState<string>(null);
```
socjalis: the correct type is string | null, unfortunately the configuration is not strict enough

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentials.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    const { width: windowWidth } = useWindowSize();
+    const rowCount = Math.max(Math.ceil((Math.abs(windowWidth - MARGIN_LIMIT)) / ACTION_MIN_WIDTH), 1);
+
+    const handleDelete = () => {
```
socjalis, line 53: I would think about moving logic like this to async thunks or using await, as I don't know if asynchronicity is done on purpose

runbotics/runbotics-scheduler/src/database/process/process.service.ts
```diff
@@ -30,4 +31,29 @@ export class ProcessService {
     partialUpdate(process: PartialUpdateProcess) {
         return this.processRepository.update({ id: process.id }, process);
     }
+
+    async checkAccessAndGetById(processId: number, user: UserEntity) {
+        // TODO: to be updated after process migration to scheduler
+        // Explanation: Process privileges should be based on process collections
+        // and checking privileges should be adjusted
+        return user.authorities.find(authority => authority.name === Role.ROLE_TENANT_ADMIN)
```
socjalis: assigning result to a variable and `if` `else` would be more readable

runbotics/runbotics-scheduler/src/scheduler-database/credential-collection/credential-collection.service.ts
```diff
@@ -114,6 +114,25 @@ export class CredentialCollectionService {
         return collections;
     }
 
+    async findAllAccessibleForCredentials(user: IUser) {
```
socjalis: ForUser?

runbotics/runbotics-common/src/model/api/process-credential.model.ts
Aeri4a: I will have to discuss it with @bhszewczyk

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.tsx
Aeri4a: I will change all to filteredby

runbotics/runbotics-orchestrator-ui/src/src-app/credentials/getRequiredCredentialsTypesInProcess.ts
Aeri4a: I could add an optional chaining char (?) and in this case, we will get empty set. We should discuss the behavior.

runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Credentials/Credentials.state.ts
Aeri4a: Types will be discussed with @bhszewczyk

runbotics/runbotics-orchestrator-ui/src/src-app/translations/en/process/configure/credentials.ts
Aeri4a: accident, I will remove it

runbotics/runbotics-scheduler/src/database/process/process.service.ts
Aeri4a: I mean it depends on case. This will make us to every time check for it in function in which it will be used. I'm not sure if it is good.

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.controller.ts
Aeri4a: I can do it like that

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
Aeri4a: You're right, i forgot

runbotics/runbotics-scheduler/src/scheduler-database/credential-collection/credential-collection.service.ts
```diff
@@ -114,6 +114,25 @@ export class CredentialCollectionService {
...
+        const collections = await this.credentialCollectionRepository.find({
+            where: {
+                tenantId: user.tenantId,
+                credentialCollectionUser: {
+                    userId: user.id,
+                    privilegeType: PrivilegeType.WRITE,
+                },
+            },
+            relations: [...relations, 'credentials.createdBy', 'credentials.collection'],
```
socjalis, line 132: if we used the object argument instead of the array, we would have intellisense support

runbotics/runbotics-scheduler/src/scheduler-database/credential-collection/credential-collection.service.ts
```diff
@@ -114,6 +114,25 @@ export class CredentialCollectionService {
...
+        });
+
+        if (collections.length === 0) {
```
socjalis: empty array is a legitmate, correct response, no need for an error

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
```diff
@@ -74,13 +75,24 @@ export class CredentialService {
...
   async findAllAccessible(tenantId: string, user: IUser) {
-    const accessibleCollections = await this.collectionService.findAllAccessible(user);
+    const accessibleCollections = await this.collectionService.findAllAccessibleForCredentials(user);
 
     if (accessibleCollections.length === 0) {
       throw new NotFoundException('Could not find any accessible collections');
     }
 
-    const credentials = accessibleCollections.map((collection) => collection.credentials).flat();
+    const credentials = accessibleCollections.map((collection) => collection.credentials)
```
socjalis: it's always safer to provide param names yourself instead of using spread operator

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
```diff
@@ -89,6 +101,42 @@ export class CredentialService {
...
+        'template.name = :templateName', { templateName }
+      ).leftJoin(
+        ProcessCredential, 'pc', 'pc.credentialId = credentials.id',
+      ).where(
+        'pc.processId != :processId OR pc.processId IS NULL',
+        { processId }
+      ).getMany().then(credentials => credentials.map(credential => ({
+        ...credential,
+        createdBy: {
+          id: credential.createdBy.id,
```
socjalis: why is it necessary to map it like this?

runbotics/runbotics-scheduler/src/scheduler-database/process-credential/process-credential.controller.ts
```diff
@@ -0,0 +1,45 @@
...
+import { CreateProcessCredentialDto, createProcessCredentialSchema } from './dto/create-process-credential.dto';
+
+
+@UseInterceptors(TenantInterceptor)
+@Controller('api/scheduler/tenants/:tenantId/process-credentials')
+export class ProcessCredentialController {
+    private readonly logger = new Logger(ProcessCredentialController.name);
+    constructor(private readonly processCredentialService: ProcessCredentialService) {}
+
+    @Get('processes/:processId')
```
socjalis, line 18: this endpoint belongs to process controller: `processes/:processId/credentials`,  `@Get('processes/:processId')` suggests it returns a process

runbotics/runbotics-scheduler/src/scheduler-database/process-credential/process-credential.module.ts
```diff
@@ -0,0 +1,16 @@
+import { Module } from '@nestjs/common';
+import { TypeOrmModule } from '@nestjs/typeorm';
+import { ProcessCredential } from './process-credential.entity';
+import { ProcessCredentialController } from './process-credential.controller';
+import { ProcessCredentialService } from './process-credential.service';
+import { ProcessModule } from '#/database/process/process.module';
+import { Credential } from '../credential/credential.entity';
+
+
```
socjalis: - \n

runbotics/runbotics-scheduler/src/scheduler-database/process-credential/process-credential.service.ts
```diff
@@ -0,0 +1,129 @@
...
+        });
+
+        const processId = processCredential.process.id;
+
+        await this.processService.checkAccessAndGetById(
+            processId, user
+        );
+
+        await this.processCredentialRepository.delete(id);
+
```
socjalis, line 91: 1. this should be done in a transaction
2. 
```
const credentialsToSave = await repo.getProcess(id);
credentialsToSave.sort(a,b => a.order - b.order).filter(c => c.order < x).map((c, idx) => {...c, order: idx))
repo.update(credentialsToSave);
```

simpler code, worth the small overhead of fetching imo

runbotics/runbotics-orchestrator-ui/src/src-app/Actions/asana.actions.ts
Aeri4a: All right

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.styles.ts
Aeri4a: I will wrap it a little bit

runbotics/runbotics-scheduler/src/migrations/1724064431315-ProcessCredentialEntity.ts
Aeri4a: auto-generated

runbotics/runbotics-orchestrator-ui/src/src-app/utils/types/page.ts
Aeri4a: If we would like the URLBuilder to build url for us with custom query parameters. URLBuilder has had this functionality but the interface wasn't adjusted before.

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.tsx
Aeri4a: If find function in this component will not find anything it will return undefined and then this handle will contain undefined value. But I will simplify a bit

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.tsx
Aeri4a: Because we don't need author duplicates in select input

Aeri4a: This is MenuItem picking author, after selecting it then credentialsFilteredByAuthor will change and map for us another select input - and get now only collections from this reduced set

runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Credentials/Credentials.slice.ts
Aeri4a: it is

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentials.utils.ts
Aeri4a: Without it we would get 'any' type and make push on this 'any' type.
I could add ActionCredentialType in type of template name but I'm not sure. This is something to discuss.

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentialsAddDialog.tsx
Aeri4a: Already removed

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.tsx
Aeri4a: All right

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentialsAddDialog.tsx
Aeri4a: I will remove it then.

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentialsAddDialog.tsx
Aeri4a: I will move handleClose to callbacks in this case.

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentials.tsx
Aeri4a: Do you mean rowCount which is not async, or handleDelete?

runbotics/runbotics-scheduler/src/database/process/process.service.ts
Aeri4a: I forgot that there is actually a function for it, so I think I will use also if as you say so

runbotics/runbotics-scheduler/src/scheduler-database/credential-collection/credential-collection.service.ts
Aeri4a: It is used in credential service and I needed additional relations

runbotics/runbotics-scheduler/src/database/process/process.service.ts
Aeri4a: In case finding one, as in Java we did throw an error

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
Aeri4a: That is true, I think there is a need for DTO mapper in this service and this should be added as a refactor for all of those services

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
Aeri4a: Yes, we shouldn't for example send whole user as well as credential collection because we don't use the full object on frontend.

runbotics/runbotics-scheduler/src/scheduler-database/process-credential/process-credential.controller.ts
Aeri4a: No, it belongs to process-credentials resource and we are sorting by processId. It could be also done as a filter which we don't have already.

runbotics/runbotics-scheduler/src/scheduler-database/process-credential/process-credential.service.ts
Aeri4a: I had a feeling that transaction would fit here. I will think about those.

runbotics/runbotics-common/src/model/api/credential-collection.model.ts
```diff
@@ -0,0 +1,20 @@
...
+    description?: string;
+    createdBy?: UserDTO;
+    createdById?: string;
+    updatedBy?: UserDTO;
+    updatedById?: string;
+    credentials?: Credential[];
+    color?: string;
+    accessType?: string;
+    createdAt?: string;
+    updatedAt?: string;
```
star-in-the-jar: why newly created types have all of theirs properties optional? cannot we keep'em required just like in the entity and define DTOs types using TS Pick<> and Omit<>?

runbotics/runbotics-common/src/model/api/credential-template.model.ts
```diff
@@ -0,0 +1,6 @@
+
+export interface CredentialTemplate {
+    id?: string;
+    name?: string;
+    description?: string;
+}
```
star-in-the-jar: as above

runbotics/runbotics-common/src/model/api/credential.model.ts
```diff
@@ -0,0 +1,22 @@
...
+    collection?: CredentialCollection;
+    createdById?: string;
+    createdBy?: UserDTO;
+    updatedBy?: UserDTO;
+    updatedById?: string;
+    templateId?: string;
+    template?: CredentialTemplate;
+    createdAt?: string;
+    updatedAt?: string;
+};
```
star-in-the-jar: like before

runbotics/runbotics-orchestrator-ui/src/src-app/Actions/jira-server.actions.ts
```diff
@@ -1,11 +1,10 @@
 import { JiraServerAction, ActionRegex } from 'runbotics-common';
 
+import { ActionCredentialType } from '#src-app/credentials/actionCredentialType.enum';
```
star-in-the-jar: I've merged PR with ActionCredentialType moved to runbotics-common (cause it's being used both on scheduler and ui). Plase import it from there and remove definition from #src-app

runbotics/runbotics-orchestrator-ui/src/src-app/credentials/useRequiredCredentialTypes.ts
```diff
@@ -0,0 +1,15 @@
+import { useMemo } from 'react';
+
+import { getRequiredCredentialsTypesInProcess } from '#src-app/credentials/getRequiredCredentialsTypesInProcess';
+import { useSelector } from '#src-app/store';
+import { processSelector } from '#src-app/store/slices/Process';
+
+export const useRequiredCredentialTypes = () => {
```
star-in-the-jar: return type (unless it's its deducted by TS) should be added

runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Process/Process.state.ts
```diff
@@ -64,6 +65,12 @@ export interface UpdateDiagramRequest {
     executionInfo?: string | null;
 }
 
+export interface CreateProcessCredentialDto {
+    credentialId: string;
+    templateName: string;
+    processId: string;
+}
```
star-in-the-jar: If it has common fields with ProcessCredential type defined in rb-common it could extend/Omit/Pick properties from ProcessCredential

runbotics/runbotics-orchestrator-ui/src/src-app/translations/en/process/configure/credentials.ts
```diff
@@ -0,0 +1,14 @@
...
+    'Process.Configure.Credentials.Section.Title': 'Credentials',
+    'Process.Configure.Credentials.Action.PrimaryCredential': 'Primary',
+    'Process.Configure.Credentials.Modal.Delete.Title': 'Delete process credential',
+    'Process.Configure.Credentials.Modal.Delete.Content': 'Are you sure you want to delete process credential?',
+    'Process.Configure.Credentials.Modal.Delete.Info.Success': 'Successfully deleted process credential',
+    'Process.Configure.Credentials.Modal.Delete.Info.Error': 'Unexpected error occurred while deleting process credential',
+    'Process.Configure.Credentials.Modal.Add.Title': 'Add new credential for process action',
+    'Process.Configure.Credentials.Modal.Add.Button.Add': 'Add',
+    'Process.Configure.Credentials.Modal.Add.Info.Success': 'Successfully added process credential',
+    'Process.Configure.Credentials.Modal.Add.Info.Error': 'Unexpected error occurred while adding process credential',
```
star-in-the-jar, line 12: ✋

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentials.tsx
```diff
@@ -0,0 +1,150 @@
...
+import { ActionSortedColumns, sortByActionCredentialType, sortByColumns } from './ProcessCredentials.utils';
+import { ProcessCredentialsAddDialog } from './ProcessCredentialsAddDialog';
+import { ProcessCredentialsDeleteDialog } from './ProcessCredentialsDeleteDialog';
+
+const ACTION_MIN_WIDTH = 400;
+const MARGIN_LIMIT = 800;
+
+const ProcessCredentials = () => {
+    const { enqueueSnackbar } = useSnackbar();
+    const { translate }= useTranslations();
```
star-in-the-jar: missing space - formatting

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentials.tsx
```diff
@@ -0,0 +1,150 @@
...
+    const credentialTypes = useRequiredCredentialTypes();
+    const actionCredentials = useMemo(
+        () => credentialTypes ? sortByActionCredentialType(processCredentials, credentialTypes) : null,
+        [processCredentials, credentialTypes]
+    );
+
+    const [columns, setColumns] = useState<ActionSortedColumns>([]);
+    const [isAddDialogOpen, setIsAddDialogOpen] = useState(false);
+    const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false);
+    const [currentCredentialId, setCurrentCredentialId] = useState<string | null>(null);
```
star-in-the-jar: I think there should be created type for CredentialId = string | null. Missing it is an antipattern of primitive obsession. There's a risk that if we change it to string | null | undefined (which I think can happen) there'll be a problem with changing it across the whole project

runbotics/runbotics-scheduler/src/database/process/process.service.ts
```diff
@@ -30,4 +32,34 @@ export class ProcessService {
     partialUpdate(process: PartialUpdateProcess) {
         return this.processRepository.update({ id: process.id }, process);
     }
+
+    async checkAccessAndGetById(processId: number, user: UserEntity) {
+        // TODO: to be updated after process migration to scheduler
```
star-in-the-jar, line 37: we've settled to allow for running the process with credentials if user does have access to the process, so this todo is not necessary after all. @socjalis correct me if I am mistaken. 

Sorry for not pointing this out in FigJam, I'll do it

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
```diff
@@ -72,9 +73,24 @@ export class CredentialService {
   }
 
   async findAllAccessible(user: IUser) {
-    const accessibleCollections = await this.collectionService.findAllAccessible(user);
+    const accessibleCollections = await this.collectionService.findAllAccessibleWithUser(user);
 
-    const credentials = accessibleCollections.map((collection) => collection.credentials).flat();
+    if (accessibleCollections.length === 0) {
+      throw new NotFoundException('Could not find any accessible collections');
+    }
```
star-in-the-jar: should we really throw an error or return empty array and on the front if array is empty display not error but UI message "No collections found"?

runbotics/runbotics-common/src/model/api/credential-collection.model.ts
Aeri4a: Sure, did it

runbotics/runbotics-orchestrator-ui/src/src-app/credentials/useRequiredCredentialTypes.ts
Aeri4a: added

runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Process/Process.state.ts
Aeri4a: done

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentials.tsx
Aeri4a: I will do it

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
Aeri4a: I could return here empty array

runbotics/runbotics-scheduler/src/database/process/process.service.ts
Aeri4a: "if user does have access to the process" - and this function is actually checking this, but how u can see it represents the old approach without process collections check. To check privilege to process we need to check privilege to its collection and all above.

In case of UI probably it doesn't have to be checked, but in API definitely imo.

marcin-depcik (approved):
fix build

runbotics/runbotics-common/src/model/api/credential-collection.model.ts
```diff
@@ -0,0 +1,20 @@
...
+    createdBy: UserDTO;
+    createdById: string;
+    updatedBy: UserDTO;
+    updatedById: string;
+    credentials: Credential[];
+    color: string;
+    accessType: string;
+    createdAt: string;
+    updatedAt: string;
+}
```
star-in-the-jar, line 20: image

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+    const [author, setAuthor] = useState<string>(null);
+    const credentialsFilteredByAuthor = credentials.filter(
+        (credential) => credential.createdBy.login === author
+    );
+
+    const [collectionName, setCollectionName] = useState<string>(null);
+    const credentialsFilteredByCollection = credentialsFilteredByAuthor.filter(
+        (credential) => credential.collection.name === collectionName
+    );
```
star-in-the-jar, line 30: maybe useMemo

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentialsAddDialog.tsx
```diff
@@ -0,0 +1,103 @@
...
+                <DialogTitle>
+                    <Typography variant='h5'>
+                        {translate('Process.Configure.Credentials.Modal.Add.Title')}
+                    </Typography>
+                </DialogTitle>
+                <AddDialogContent>
+                    <CredentialSelectSet
+                        credentials={all ?? []}
+                        handleCredentialChange={(credential) => { setPickedCredential(credential); }}
+                    />
```
star-in-the-jar: pass function in useCallback

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.tsx
```diff
@@ -0,0 +1,108 @@
...
+                >
+                    {_.uniqBy(credentials, credential => credential.createdBy.login)
+                        .map(credential => (
+                            <MenuItem
+                                value={credential.createdBy.login}
+                                key={credential.createdBy.login}
+                            >
+                                {credential.createdBy.login}
+                            </MenuItem>
+                        ))}
```
star-in-the-jar: consider extracting to const variable for greater readability

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.tsx
```diff
@@ -0,0 +1,108 @@
...
+                >
+                    {_.uniqBy(credentialsFilteredByAuthor, credential => credential.collection.name)
+                        .map(credential => (
+                            <MenuItem
+                                value={credential.collection.name}
+                                key={credential.collection.name}
+                            >
+                                {credential.collection.name}
+                            </MenuItem>
+                        ))}
```
star-in-the-jar: consider moving to named const var

runbotics/runbotics-orchestrator-ui/src/src-app/credentials/getRequiredCredentialsTypesInProcess.ts
```diff
@@ -0,0 +1,14 @@
...
+const ACTION_ID_REGEX = /camunda:actionId="([a-zA-Z.]+)"/g;
+
+export const getRequiredCredentialsTypesInProcess = (definition: IProcess['definition']): ActionCredentialType[] =>
+    [
+        ...Array.from(definition.matchAll(ACTION_ID_REGEX), m => m[1])
+            .reduce((set, actionId) =>
+                actions[actionId]?.credentialType
+                    ? set.add(actions[actionId].credentialType) : set
+            , new Set<ActionCredentialType>())
+    ];
```
star-in-the-jar, line 14: why do you need [...] if you have Array.from?

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentials.tsx
```diff
@@ -0,0 +1,154 @@
...
+                handleClose={handleDeleteDialogClose}
+                handleDelete={handleDelete}
+            />
+            <ProcessCredentialsAddDialog
+                isOpen={isAddDialogOpen}
+                handleClose={handleAddDialogClose}
+                templateName={currentActionName}
+            />
+            <Header>
+                <Image src={LockIcon} alt='lock icon' style={{ filter: 'brightness(0) saturate(100%)' }}/>
```
star-in-the-jar: alts should be translated too

star-in-the-jar: it's all right here but because of lower priority it's preferred to use styled components instead of style prop.

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentials.utils.ts
star-in-the-jar: template.name should be type of ActionCredentialType. If needed add migration that changes this field type and adds enum to column.

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentialsAddDialog.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <If condition={isOpen}>
+            <Dialog open fullWidth>
+                <DialogTitle>
+                    <Typography variant='h5'>
+                        {translate('Process.Configure.Credentials.Modal.Add.Title')}
+                    </Typography>
+                </DialogTitle>
+                <AddDialogContent>
+                    <CredentialSelectSet
+                        credentials={all ?? []}
```
star-in-the-jar: if there are no credential does the component display 'no credentials found' message?

star-in-the-jar (approved):

star-in-the-jar:
version to bumb

runbotics/runbotics-orchestrator-ui/src/src-app/components/CredentialSelectSet/CredentialSelectSet.tsx
Aeri4a: I used it, @socjalis said that is is not necessary so I changed it

runbotics/runbotics-orchestrator-ui/src/src-app/credentials/getRequiredCredentialsTypesInProcess.ts
Aeri4a: cuz it will return Set as you can see from reduce function

runbotics/runbotics-orchestrator-ui/src/src-app/views/process/ProcessConfigureView/ProcessCredentials/ProcessCredentialsAddDialog.tsx
Aeri4a: It will be also added before merge.

PR #754, opened 2024-10-01
[fix/RPA-1725] Set cells invalid input type

marcin-depcik:
## Types of changes

---


runbotics/runbotics-desktop/src/action/google/google.action-handler.ts
```diff
@@ -187,4 +188,17 @@ export default class GoogleActionHandler extends StatelessActionHandler {
...
+
+            if (Array.isArray(values)) {
+                return values;
+            }
+
+            return JSON.parse(values);
+        } catch (e) {
+            throw new Error('Invalid values type. Should be 2D array.');
+        }
+    }
```
bhszewczyk, line 203: if this is possible you could get rid of unknown :)

bhszewczyk (approved):

runbotics/runbotics-desktop/src/action/google/google.action-handler.ts
```diff
@@ -187,4 +188,17 @@ export default class GoogleActionHandler extends StatelessActionHandler {
             values: [],
         };
     }
+    private inputArrayParser(values: unknown[][] | string): unknown[][] {
```
star-in-the-jar: in the schema you specified values can be: string[] | (string | number)[][] so you should create a type for that

star-in-the-jar (approved):

PR #773, opened 2024-10-11
[RPA-1703.2] Credential change email notification - email update

Aeri4a:
…cise edit operation info

## Types of changes

---

### DIFF
--- a/runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
+++ b/runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
@@ -71,6 +71,14 @@ export class CredentialService {
     });
 
     return this.credentialRepo.save(credential)
+      .then((credential) => {
+        this.notifyCredentialCollectionOwner({
+          executor: user,
+          collectionId: credential.collectionId,
+          credentialName: credential.name,
+          operationType: CredentialOperationType.CREATE,
+      });
+      })
       .catch(async (error) => {
         await this.validateName(credentialDto.name, collectionId, tenantId);
         throw new BadRequestException(error.message);
@@ -222,6 +230,9 @@ export class CredentialService {
       tenantId: credential.tenantId,
     });
 
+    const credentialOldName =
+        credential.name !== credentialDto.name ? credential.name : undefined;
+
     return this.credentialRepo
         .save(credentialToUpdate)
         .then((credential) => {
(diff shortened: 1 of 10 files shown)


runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.utils.ts
```diff
@@ -23,12 +24,14 @@ interface BaseCredentialChangeMailPayload {
     collectionCreatorEmail: string;
     collectionName: string;
     credentialName: string;
+    credentialOldName?: string | undefined;
```
star-in-the-jar: If undefined is allowed, and this is the only use of the interface, why to make credentialOldName optional?

star-in-the-jar (approved):
+ version

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
```diff
@@ -71,6 +71,14 @@ export class CredentialService {
...
 
     return this.credentialRepo.save(credential)
+      .then((credential) => {
+        this.notifyCredentialCollectionOwner({
+          executor: user,
+          collectionId: credential.collectionId,
+          credentialName: credential.name,
+          operationType: CredentialOperationType.CREATE,
+      });
+      })
```
marcin-depcik: can you lint this?

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.utils.ts
```diff
@@ -23,12 +24,14 @@ interface BaseCredentialChangeMailPayload {
...
 }
 
 interface BaseCredentialNotifyMailArgs {
     executor: IUser;
     collectionId: string;
     credentialName: string;
+    credentialOldName?: string | undefined;
```
marcin-depcik, line 34: I think you should decide if you want to require this property, and if you do, passed value can be only string or undefined, or if you want to set this parameter as optional and always require a string if this property is defined

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.service.ts
```diff
@@ -222,13 +230,17 @@ export class CredentialService {
       tenantId: credential.tenantId,
     });
 
+    const credentialOldName =
+        credential.name !== credentialDto.name ? credential.name : undefined;
```
marcin-depcik, line 235: this should be adjusted depending on what approach you're gonna take in the interfaces

marcin-depcik (approved):

runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.utils.ts
Aeri4a: You're right

PR #817, opened 2024-11-22
[RPA-1819] Guest migration

Aeri4a:
## Types of changes

---


runbotics/runbotics-scheduler/src/scheduler-database/guest/guest.controller.ts
```diff
@@ -0,0 +1,32 @@
...
+@Controller('/api/scheduler/tenants/:tenantId/guests')
+export class GuestController {
+    private readonly logger = new Logger(GuestController.name);
+
+    constructor(
+        private readonly guestService: GuestService,
+    ) {}
+
+    @Get('user')
+    async getGuestByUser(
```
bhszewczyk, line 19: I would add here Id (getGuestByUserId) or even leave just getGuest. I find the current name a little bit confusing

runbotics/runbotics-scheduler/src/scheduler-database/guest/guest.service.ts
```diff
@@ -0,0 +1,50 @@
...
+        );
+    }
+
+    async decrementExecutionsCount(userId: number): Promise<UpdateResult> {
+        return this.guestRepository.decrement(
+            { user: { id: userId } },
+            'executionsCount',
+            1
+        );
+    }
```
bhszewczyk, line 49: do we need this? is there any reason we would decrement execution count?

bhszewczyk (approved):

bhszewczyk:
remember to bump version :)

runbotics/runbotics-scheduler/src/scheduler-database/guest/guest.controller.ts
Aeri4a: getGuest will fit, I will change it :D

runbotics/runbotics-scheduler/src/scheduler-database/guest/guest.service.ts
Aeri4a: yes, in case something with process execution failed without user intention

runbotics/runbotics-scheduler/src/migrations/1732182583076-Guests.ts
```diff
@@ -0,0 +1,18 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class Guests1732182583076 implements MigrationInterface {
+    name = 'Guests1732182583076'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "guest" DROP CONSTRAINT "fk_guest_user_id"`);
+        await queryRunner.query(`ALTER TABLE "guest" ALTER COLUMN "user_id" DROP NOT NULL`);
```
marcin-depcik: is this intentional?

marcin-depcik (approved):
bump version

runbotics/runbotics-scheduler/src/migrations/1732182583076-Guests.ts
Aeri4a: It shouldn't be here, good catch. I discovered that if we don't specify column in entity and make just some kind of relation like ManyToOne, then those columns in that relation will be nullable - which is not explicitly mentioned.

PR #819, opened 2024-11-25
[RPA-1792] Fixes bug in schedule process cron select

bhszewczyk:
## Description

Before:
- default schedule value was minute
- once other value that minute was selected other inputs where not mandatory and after submitting the schedule was set to run every minute

After:
- changed default schedule cron value from minute to day
- schedule inputs are now required for all options other than minute

## Types of changes

---

### DIFF
--- a/runbotics/runbotics-orchestrator-ui/src/src-app/components/cron/components/CustomSelect/CustomSelect.tsx
+++ b/runbotics/runbotics-orchestrator-ui/src/src-app/components/cron/components/CustomSelect/CustomSelect.tsx
@@ -22,15 +22,15 @@ const CustomSelect: FC<CustomSelectProps> = ({
     isMultiple = true,
 }) => {
     const stringValue = useMemo(
-        () => 
-            (value && Array.isArray(value)) 
+        () =>
+            (value && Array.isArray(value))
                 ? value.map((item: number) => item.toString())
                 : [],
         [value]
     );
 
     const options = useMemo(
-        () => optionsList 
+        () => optionsList
             ? (
                 optionsList.map(
                     (option, index) => {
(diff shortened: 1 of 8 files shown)


runbotics/runbotics-orchestrator-ui/src/src-app/components/cron/components/CustomSelect/CustomSelect.tsx
```diff
@@ -109,6 +109,8 @@ const CustomSelect: FC<CustomSelectProps> = ({
             renderValue={renderTag}
             autoWidth={false}
             disabled={disabled}
+            required
+            error={!value.length}
```
star-in-the-jar, line 113: is there appropriate error message?

star-in-the-jar (approved):
+ version

runbotics/runbotics-orchestrator-ui/src/src-app/components/cron/components/CustomSelect/CustomSelect.tsx
star-in-the-jar: user has to now what did he/she done wrong to make it right. When that's added feel free to merge

bhszewczyk: Consulted this with Ada and agreed that there is no need for this in this particular case - it is handled by browser after submit that this field is required and as long as it it empty the borders are red. If that is truly necessary let me know.

star-in-the-jar: If there are red borders around empty inputs it's correct

Aeri4a (approved):

PR #839, opened 2024-12-10
[RPA-1857] Add migration for old logs management.

star-in-the-jar:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-scheduler/src/migrations/1733815803333-OldLogsManagement.ts
+++ b/runbotics/runbotics-scheduler/src/migrations/1733815803333-OldLogsManagement.ts
@@ -1,0 +1,149 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class OldLogsManagement1733815803333 implements MigrationInterface {
+    name = 'OldLogsManagement1733815803333';
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`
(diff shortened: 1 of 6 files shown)


runbotics/runbotics-scheduler/src/migrations/1733815803333-OldLogsManagement.ts
```diff
@@ -0,0 +1,149 @@
...
+                        WHERE created < cutoff_date
+                            AND process_id IN (
+                                SELECT process_id
+                                FROM process_instance
+                                WHERE created > half_year_back
+                                GROUP BY process_id
+                                HAVING COUNT(id) > 10
+                            )
+                        LIMIT batch_limit
+                    );
```
marcin-depcik, line 40: what is the difference between `cutoff_date` and `half_year_back`?

runbotics/runbotics-scheduler/src/migrations/1733815803333-OldLogsManagement.ts
marcin-depcik, line 1: grate job!

marcin-depcik (approved):

runbotics/runbotics-scheduler/src/migrations/1733815803333-OldLogsManagement.ts
star-in-the-jar: `cutoff_date` is given by the user and is `<today_minus_6_months>` by default, but can be set to any date. Logs after that date are supposed to be deleted.
`half_year_back` is always `<today_minus_6_months>`, is used to check how many runs the process does have in the last half year

PR #847, opened 2024-12-12
[RPA-1890] Change api/processes to scheduler

Aeri4a:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-desktop/src/action/general/general.action-handler.ts
+++ b/runbotics/runbotics-desktop/src/action/general/general.action-handler.ts
@@ -4,7 +4,7 @@ import { GeneralAction, IProcess, ProcessInstanceStatus, ITriggerEvent, BotSyste
 import { delay } from '#utils';
 import { RunboticsLogger } from '#logger';
 import { RuntimeService } from '#core/bpm/runtime';
-import { orchestratorAxios } from '#config';
+import { schedulerAxios, StorageService } from '#config';
 import getBotSystem from '#utils/botSystem';
 import { ConsoleLogActionInput, ConsoleLogActionOutput, DelayActionInput, DelayActionOutput, StartProcessActionInput, StartProcessActionOutput, ThrowErrorActionInput, ThrowErrorActionOutput } from './general.types';
 
@@ -21,6 +21,7 @@ export default class GeneralActionHandler extends StatelessActionHandler {
     constructor(
         @Inject(forwardRef(() => RuntimeService))
         private runtimeService: RuntimeService,
+        private readonly storageService: StorageService,
     ) {
         super();
     }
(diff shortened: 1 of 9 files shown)


runbotics/runbotics-desktop/src/action/general/general.action-handler.ts
```diff
@@ -21,6 +21,7 @@ export default class GeneralActionHandler extends StatelessActionHandler {
     constructor(
         @Inject(forwardRef(() => RuntimeService))
         private runtimeService: RuntimeService,
+        private readonly storageService: StorageService,
```
marcin-depcik, line 24: I think you need to import ConfigModule inside CoreModule

runbotics/runbotics-desktop/src/action/general/general.action-handler.ts
```diff
@@ -44,8 +45,9 @@ export default class GeneralActionHandler extends StatelessActionHandler {
     async startProcess(
         request: DesktopRunRequest<GeneralAction.START_PROCESS, StartProcessActionInput>
     ): Promise<StartProcessActionOutput> {
-        const response = await orchestratorAxios.get<IProcess>(
-            `/api/processes/${request.input.processId}`,
+        const tenantId = this.storageService.getValue('tenantId');
```
marcin-depcik, line 48: I know that, we have such implementation, but I don't think this is a good approach of keeping such information like `tenantId`. Maybe we should think about changing it in the future

marcin-depcik (approved):

marcin-depcik (commented):
bump version

runbotics/runbotics-desktop/src/action/general/general.action-handler.ts
star-in-the-jar: Good point, if it's stored in local storage there's a risk someone can exploit that to access resources of another tenant

runbotics/runbotics-desktop/src/core/websocket/listeners/process.listener.ts
```diff
@@ -89,8 +90,11 @@ export class ProcessListener {
             }
 
             const { processId, input, ...rest } = data;
-            const process = await orchestratorAxios
-                .get<IProcess>(`/api/processes/${processId}`, { maxRedirects: 0 })
+            const tenantId = this.storageService.getValue('tenantId');
```
star-in-the-jar, line 93: If there's a risk of overwriting this value by a user, we need a bugfix task in #tobediscussed

star-in-the-jar (changes requested):
+version

runbotics/runbotics-desktop/src/core/websocket/listeners/process.listener.ts
marcin-depcik: I created task for this

PR #849, opened 2024-12-12
[RPA-1483] Check and update resources by tenantId

marcin-depcik:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-scheduler/src/queue/queue.service.ts
+++ b/runbotics/runbotics-scheduler/src/queue/queue.service.ts
@@ -170,12 +170,12 @@ export class QueueService implements OnModuleInit {
 
     async validateProcessAccess({ process, user, triggered = false }: ValidateProcessAccessProps) {
         const hasTenantAccess = process.tenantId === user.tenantId;
-        const hasAccess = hasTenantAccess && process.createdBy?.id === user?.id;
+        const hasProcessAccess = hasTenantAccess && process.createdBy?.id === user?.id;
         const isPublic = hasTenantAccess && process.isPublic;
-        const isAdmin = hasTenantAccess && user?.authorities.filter(role => role.name === Role.ROLE_ADMIN || role.name === Role.ROLE_TENANT_ADMIN).length > 0;
+        const isAdmin = hasTenantAccess && user?.authorities.filter(role => [Role.ROLE_ADMIN, Role.ROLE_TENANT_ADMIN].includes(role.name)).length > 0;
         const isTriggerable = process?.isTriggerable;
 
-        if (!hasAccess && !isPublic && !isAdmin) {
+        if (!hasProcessAccess && !isPublic && !isAdmin) {
             this.logger.error(`User${user ? ' ' + user?.email : ''} does not have access to the process "${process?.id}"`);
             throw new ForbiddenException(`You do not have access to the process "${process?.id}"`);
         }
--- a/runbotics/runbotics-scheduler/src/queue/queue.service.ts
+++ b/runbotics/runbotics-scheduler/src/queue/queue.service.ts
@@ -169,10 +169,11 @@ export class QueueService implements OnModuleInit {
     }
 
     async validateProcessAccess({ process, user, triggered = false }: ValidateProcessAccessProps) {
-        const hasAccess = process.createdBy?.id === user?.id;
-        const isPublic = process.isPublic;
+        const hasTenantAccess = process.tenantId === user.tenantId;
+        const hasAccess = hasTenantAccess && process.createdBy?.id === user?.id;
+        const isPublic = hasTenantAccess && process.isPublic;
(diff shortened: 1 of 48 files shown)


runbotics/runbotics-scheduler/src/scheduler-database/process-instance/process-instance.controller.ts
```diff
@@ -36,21 +36,24 @@ export class ProcessInstanceController {
...
+    ) {
         return this.processInstanceService.getOne(id, user);
     }
 
     @Get(':id/subprocesses/GetPage')
     getSubprocesses(
         @Param('id') id: ProcessInstance['id'],
         @Pageable() paging: Paging,
-        @UserDecorator() user: User
+        @UserDecorator() user: User,
```
star-in-the-jar, line 56: do you have some extension or is it ESlint?

star-in-the-jar: that adds comma

runbotics/runbotics-scheduler/src/scheduler-database/tags/tag.service.ts
```diff
@@ -16,7 +16,7 @@ export class TagService {
         private readonly tagRepository: Repository<Tag>
     ) {}
 
-    async getAll(tenantId: string, searchPhrase?: string) {
+    async getAll(tenantId: string, searchPhrase: string | undefined) {
```
star-in-the-jar: what is the reason? shouldn't it stay as it was?

star-in-the-jar (approved):

runbotics/runbotics-scheduler/src/scheduler-database/process-instance/process-instance.controller.ts
marcin-depcik: indeed, my hand is an extension :D

runbotics/runbotics-scheduler/src/scheduler-database/tags/tag.service.ts
marcin-depcik: so far I know, `?` means that argument is optional, so it is unnecessary to pass to the function, but `string | undefined` means that argument must be passed but can be undefined

runbotics/runbotics-scheduler/src/scheduler-database/process-instance/process-instance.controller.ts
marcin-depcik: but I think you can configure prettier in the same way

runbotics/runbotics-scheduler/src/queue/queue.service.ts
```diff
@@ -169,10 +169,11 @@ export class QueueService implements OnModuleInit {
     }
 
     async validateProcessAccess({ process, user, triggered = false }: ValidateProcessAccessProps) {
-        const hasAccess = process.createdBy?.id === user?.id;
-        const isPublic = process.isPublic;
+        const hasTenantAccess = process.tenantId === user.tenantId;
+        const hasAccess = hasTenantAccess && process.createdBy?.id === user?.id;
```
bhszewczyk: what do you think about renaming this to hasProcessAccess?

runbotics/runbotics-scheduler/src/scheduler-database/process-instance/process-instance.controller.ts
star-in-the-jar: woooooow, nice
image

runbotics/runbotics-scheduler/src/scheduler-database/tags/tag.service.ts
star-in-the-jar: correct, I just wonder why is it required

star-in-the-jar: I think we should pass it to the function only if it is not undefined

runbotics/runbotics-scheduler/src/scheduler-database/tenant/tenant.controller.ts
```diff
@@ -34,11 +34,13 @@ export class TenantController {
 
     @GetWithTenant('me')
     @FeatureKeys(FeatureKey.TENANT_READ)
-    async getTenantByUser(@UserDecorator('tenantId') id: Tenant['id']) {
-        const tenant = await this.tenantService.getById(id);
+    async getTenantByUser(
+        @UserDecorator('tenantId') tenantId: User['tenantId'],
```
bhszewczyk: you could make these two consistent 1. @UserDecorator('tenantId') tenantId: User['tenantId']  2. @UserDecorator() { tenantId }: User
personally, I prefer @UserDecorator() { tenantId }: User,

bhszewczyk (approved):

PR #873, opened 2025-01-10
[RPA-1871] Add credentials support for desktop actions

marcin-depcik:
## Types of changes

---


runbotics/runbotics-desktop/src/action/desktop/desktop.action-handler.ts
```diff
@@ -296,11 +307,15 @@ export default class DesktopActionHandler extends StatelessActionHandler {
     }
 
     run(request: DesktopActionRequest) {
+        const credential = credentialAttributesMapper<DesktopCredential>(request.credentials);
```
bhszewczyk: I think you can move it to the case: TYPE_CREDENTIALS, only there it is needed

bhszewczyk (approved):

PR #878, opened 2025-01-15
[RPA-1877] Add decline user functionality

marcin-depcik:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-common/src/model/api/feature-key.model.ts
+++ b/runbotics/runbotics-common/src/model/api/feature-key.model.ts
@@ -212,6 +212,9 @@ export enum FeatureKey {
     // Allows to edit users assigned to the tenant
     TENANT_EDIT_USER = 'TENANT_EDIT_USER',
 
+    // Allows to delete users assigned to the tenant
+    TENANT_DELETE_USER = 'TENANT_DELETE_USER',
+
     // Allows to get invite code for admin of specific tenant
     TENANT_GET_INVITE_CODE = 'TENANT_GET_INVITE_CODE',
 
--- a/runbotics/runbotics-common/src/model/api/feature-key.model.ts
+++ b/runbotics/runbotics-common/src/model/api/feature-key.model.ts
@@ -212,8 +212,8 @@ export enum FeatureKey {
     // Allows to edit users assigned to the tenant
     TENANT_EDIT_USER = 'TENANT_EDIT_USER',
 
-    // Allows to delete users assigned to the tenant
-    TENANT_DELETE_USER = 'TENANT_DELETE_USER',
+    // Allows to decline users assigned to the tenant
+    TENANT_DECLINE_USER = 'TENANT_DECLINE_USER',
 
     // Allows to get invite code for admin of specific tenant
     TENANT_GET_INVITE_CODE = 'TENANT_GET_INVITE_CODE',
--- a/runbotics/runbotics-common/src/model/api/user.model.ts
+++ b/runbotics/runbotics-common/src/model/api/user.model.ts
@@ -11,7 +11,7 @@ export interface User {
     imageUrl: string | null;
     langKey: string;
     activated: boolean;
-    wasEverActivated: boolean;
+    hasBeenActivated: boolean;
     activationKey?: string | null;
     resetKey?: string | null;
     createdBy: string;
--- a/runbotics/runbotics-common/src/model/api/user.model.ts
+++ b/runbotics/runbotics-common/src/model/api/user.model.ts
@@ -11,6 +11,7 @@ export interface User {
     imageUrl: string | null;
     langKey: string;
     activated: boolean;
+    wasEverActivated: boolean;
     activationKey?: string | null;
     resetKey?: string | null;
     createdBy: string;
--- a/runbotics/runbotics-orchestrator-ui/src/src-app/views/users/DeclineUserReasonSelector/DeclineUserReasonSelector.tsx
+++ b/runbotics/runbotics-orchestrator-ui/src/src-app/views/users/DeclineUserReasonSelector/DeclineUserReasonSelector.tsx
@@ -100,6 +100,7 @@ export const DeclineUserReasonSelector = ({
                     rows={4}
                     value={customMessage}
                     onChange={handleTextFiledChange}
+                    disabled={reason !== DeclineReason.THIRD_REASON}
                 />
             </StyledRadioGroup>
         </StyledFormControl>
--- a/runbotics/runbotics-orchestrator-ui/src/src-app/views/users/DeclineUserReasonSelector/DeclineUserReasonSelector.tsx
+++ b/runbotics/runbotics-orchestrator-ui/src/src-app/views/users/DeclineUserReasonSelector/DeclineUserReasonSelector.tsx
@@ -50,7 +50,6 @@ export const DeclineUserReasonSelector = ({
     ) => {
         const inputValue = (event.target as HTMLInputElement).value;
 
-        setReason(inputValue);
         if (
             reason !== DeclineReason.THIRD_REASON &&
             inputValue === DeclineReason.THIRD_REASON
@@ -59,6 +58,7 @@ export const DeclineUserReasonSelector = ({
         } else {
             onMessageChange(declineReason[inputValue].message);
         }
+        setReason(inputValue);
     };
 
     const handleTextFiledChange = (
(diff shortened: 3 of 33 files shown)


runbotics/runbotics-common/src/model/api/user.model.ts
```diff
@@ -11,6 +11,7 @@ export interface User {
     imageUrl: string | null;
     langKey: string;
     activated: boolean;
+    wasEverActivated: boolean;
```
star-in-the-jar: hasBeenActivated

runbotics/runbotics-orchestrator-ui/src/src-app/views/users/DeleteUserDialog.tsx
```diff
@@ -45,24 +46,39 @@ const DeleteUserDialog: FC<DeleteUserDialogProps> = ({
...
         Promise.allSettled(
-            usersData.map((user) =>
-                dispatch(usersActions.deleteUser({ id: user.id }))
-            )
+            usersData.map((user) => hasTenantAdminAccess
+                ? dispatch(usersActions.deleteUserInTenant({
+                    resourceId: user.id,
+                    payload: {
+                        data: {
+                            declineReason: `No consent by (tenant-admin@localhost) for user (${user.email}) activation`,
```
star-in-the-jar: it does not have to be tenant-admin, but other user with tenant-admin role

star-in-the-jar: I understand the whole declineReason will be specified by a user?

runbotics/runbotics-scheduler/src/mail/mail.module.ts
```diff
@@ -30,9 +29,8 @@ import { NotificationBotModule } from '#/scheduler-database/notification-bot/not
             }),
         }),
         Logger,
-        BotModule,
+        forwardRef(() => BotModule),
```
star-in-the-jar, line 32: why is that needed here?

runbotics/runbotics-scheduler/src/migrations/1736847689046-UserWasEverActivated.ts
```diff
@@ -0,0 +1,27 @@
+import { User } from '#/scheduler-database/user/user.entity';
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class UserWasEverActivated1736847689046 implements MigrationInterface {
+    name = 'UserWasEverActivated1736847689046';
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(
+            `ALTER TABLE "jhi_user" ADD "was_ever_activated" boolean NOT NULL DEFAULT false`
```
star-in-the-jar: has_been_activated

star-in-the-jar: additionaly - we need a migration that will set every existing user from NULL to true/false, don't we?

star-in-the-jar: otherwise there can be problems when updating old users

runbotics/runbotics-scheduler/src/scheduler-database/user/user.entity.ts
```diff
@@ -37,6 +37,13 @@ export class User {
     @Column({ type: 'boolean' })
     activated: boolean;
 
+    @Column({
+        name: 'was_ever_activated',
```
star-in-the-jar: has_been_activated

runbotics/runbotics-scheduler/src/scheduler-database/user/user.service.ts
```diff
@@ -122,6 +129,26 @@ export class UserService {
 
             throw new BadRequestException();
         });
+
+        this.mailService.sendUserDeclineReasonMail(userToDelete, userDto);
+    }
+
+    async deleteInTenant(id: number, user: User, userDto: DeleteUserDto) {
+        if (!isTenantAdmin(user)) {
+            throw new ForbiddenException();
```
star-in-the-jar: ForbiddenException("Only users with Tenant-admin role can decline users' activation requests")

star-in-the-jar: or 'delete users'

runbotics/runbotics-scheduler/src/scheduler-database/user/user.service.ts
```diff
@@ -122,6 +129,26 @@ export class UserService {
...
+        }
+        const userToDelete = await this.userRepository
+            .findOneByOrFail({ id, tenantId: user.tenantId })
+            .catch(() => {
+                throw new NotFoundException();
+            });
+        if (userToDelete.wasEverActivated) {
+            throw new BadRequestException('User has been activated at least once');
```
star-in-the-jar, line 144: exactly! has been xD

star-in-the-jar (changes requested):

runbotics/runbotics-orchestrator-ui/src/src-app/views/users/DeleteUserDialog.tsx
marcin-depcik: This message is just mock, in the final result this message will be created by user who declines

runbotics/runbotics-scheduler/src/mail/mail.module.ts
marcin-depcik: there was a problem with circular deps in nestjs modules after injecting MailModule into UserModule

runbotics/runbotics-scheduler/src/migrations/1736847689046-UserWasEverActivated.ts
marcin-depcik: in this case if this column is set to `NOT NULL` and `DEFAULT false`, all current users will get `false` for `has_been_activated` after migration, and then there is logic which changes these values to `true`

runbotics/runbotics-common/src/model/api/feature-key.model.ts
```diff
@@ -205,6 +205,9 @@ export enum FeatureKey {
     // Allows to edit users assigned to the tenant
     TENANT_EDIT_USER = 'TENANT_EDIT_USER',
```
bhszewczyk, line 214: I am not sure it should be named like this. on one hand now it will be allowed to decline user activation rather than to delete a user, right? now tenant admin cannot delete the user which has been activated at least once, but in the future we may implement such feature. on the other hand we may want to just refactor this feature (so in one endpoint tenant admin would delete deactivated user and decline user activation - I leave it to you, just wanted to share my thoughts on this.

bhszewczyk (approved):

runbotics/runbotics-orchestrator-ui/src/src-app/views/users/DeleteUserDialog.utils.tsx
```diff
@@ -0,0 +1,14 @@
+interface RejectionPayload {
+    payload: {
+        message: string;
+    };
+}
+
+export const hasErrorMessage = (value: unknown): value is RejectionPayload =>
+    value !== undefined &&
+    value !== null &&
+    typeof value === 'object' &&
```
bhszewczyk:
```suggestion
    (!value || typeof value !== 'object') &&
```

runbotics/runbotics-scheduler/src/mail/mail.service.ts
```diff
@@ -165,6 +166,17 @@ export class MailService {
         this.sendMail(sendMailInput);
     }
 
+    public sendUserDeclineReasonMail(userToDelete: User, userDto: DeleteUserDto) {
+        if (userDto && 'declineReason' in userDto) {
```
bhszewczyk, line 170: As a user who awaits acceptation I would expect to receive an email regarding rejection even though there was no reason selected with something default like "Activation of your account has been declined by the tenant admin" but if this is the indented behavior that's ok.

bhszewczyk (approved):

runbotics/runbotics-orchestrator-ui/src/src-app/views/users/DeclineUserReasonSelector/DeclineUserReasonSelector.tsx
```diff
@@ -0,0 +1,108 @@
...
+    const handleRadioControlChange = (
+        event: React.ChangeEvent<HTMLInputElement>
+    ) => {
+        const inputValue = (event.target as HTMLInputElement).value;
+
+        setReason(inputValue);
+        if (
+            reason !== DeclineReason.THIRD_REASON &&
+            inputValue === DeclineReason.THIRD_REASON
+        ) {
```
bhszewczyk: tbh I don't quite get this logic, sounds like useEffect case for me.

PR #883, opened 2025-01-27
[RPA-1920] Add credential support for custom actions

marcin-depcik:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-orchestrator-ui/src/src-app/Actions/types.ts
+++ b/runbotics/runbotics-orchestrator-ui/src/src-app/Actions/types.ts
@@ -56,7 +56,7 @@ export type IBpmnAction = {
     input?: IActionInput;
     output?: IActionOutput;
     form?: IForm;
-    credentialType?: ActionCredentialType;
+    credentialType?: ActionCredentialType | string;
 };
 
 export type FormState = {
--- a/runbotics/runbotics-orchestrator-ui/src/src-app/types/model/action.model.ts
+++ b/runbotics/runbotics-orchestrator-ui/src/src-app/types/model/action.model.ts
@@ -2,6 +2,7 @@ export interface IAction {
     id?: string;
     label?: string | null;
     script?: string | null;
+    credentialType?: string | null;
     form?: string | null;
 }
 
@@ -48,6 +49,7 @@ const defaultForm = {
 export const defaultValue: Readonly<IAction> = {
     id: '',
     label: '',
+    credentialType: '',
     script: 'external.',
     form: JSON.stringify(defaultForm, null, '\t'),
 };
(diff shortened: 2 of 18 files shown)


runbotics/runbotics-scheduler/src/scheduler-database/action/action.entity.ts
```diff
@@ -18,6 +19,9 @@ export class Action {
     @Column({ type: 'varchar', length: 255 })
     script: string;
 
+    @Column({ type: 'varchar', length: 255, default: '' })
```
bhszewczyk, line 22: I am wondering: maybe this can be nullable and as default it will be null?

runbotics/runbotics-orchestrator-ui/src/src-app/views/action/ActionDetails/ExternalActionForm.tsx
```diff
@@ -80,13 +100,25 @@ export const ExternalActionForm: FC<ExternalActionFormProps> = ({ draft, setDraf
...
+                    margin: theme => `${theme.spacing(1)} 0`
                 }}
                 value={draft.label}
                 onChange={(e: ChangeEvent<HTMLInputElement>) => handleLabelChange(e)}
                 autoComplete="off"
                 {...labelInputErrorProperties}
             />
+            <TextField
+                fullWidth
+                label={'Credential Type'}
```
bhszewczyk: I think it will be better if this was taken from translation

runbotics/runbotics-orchestrator-ui/src/src-app/types/model/action.model.ts
```diff
@@ -2,6 +2,7 @@ export interface IAction {
     id?: string;
     label?: string | null;
     script?: string | null;
+    credentialType?: string | null;
```
bhszewczyk: why not ActionCredentialType | null?

runbotics/runbotics-orchestrator-ui/src/src-app/Actions/types.ts
```diff
@@ -56,7 +56,7 @@ export type IBpmnAction = {
     input?: IActionInput;
     output?: IActionOutput;
     form?: IForm;
-    credentialType?: ActionCredentialType;
+    credentialType?: ActionCredentialType | string;
```
bhszewczyk, line 59: I am not sure why we should allow a string here, could you please let me know what was the reason you needed to add this here?

bhszewczyk (changes requested):

bhszewczyk (approved):

runbotics/runbotics-scheduler/src/scheduler-database/action/action.entity.ts
marcin-depcik: I decided to leave as is for easier type handling on UI, `null` should be converted to empty string anyway

PR #884, opened 2025-01-27
[RPA-1927] Desktop actions refactor

bhszewczyk:
…values in inputs

## Description

- now zod is used for the input validation
- actions where the coordinates were needed and only variable was accepted now accept objects as values

## Types of changes

---


runbotics/runbotics-desktop/src/action/desktop/desktop.utils.ts
```diff
@@ -0,0 +1,57 @@
...
+    mouseButton: z.nativeEnum(MouseButton),
+    doubleClick: z.boolean()
+});
+
+export const typeInputSchema = z.object({
+    text: z.string()
+});
+
+export const typeCredentialsInputSchema = z.object({
+    credentialAttribute: z.enum(['username', 'password'])
```
marcin-depcik: maybe it would be better to create actual enum for this property

runbotics/runbotics-desktop/src/action/desktop/desktop.utils.ts
marcin-depcik, line 1: whye there is so many `any`

marcin-depcik (approved):

runbotics/runbotics-desktop/src/action/desktop/desktop.utils.ts
bhszewczyk: long story short - becuase of zod doesn't work well without strict - inferred type has all fields optional and I wasn't able to make it right for interface without optionals. I will try again, but I just gave up before PR :D

runbotics/runbotics-desktop/src/action/desktop/desktop.action-handler.ts
```diff
@@ -254,30 +251,38 @@ export default class DesktopActionHandler extends StatelessActionHandler {
...
+        const testingObject = targetType === ClickTarget.POINT ? new Point(0, 0) : new Region(0, 0, 0, 0);
+
+        this.isTargetObjectValid(object, testingObject);
+    }
+
+    private isTargetObjectValid(targetObject: object, testingObject: Point | Region) {
+        const pointKeys = Object.keys(testingObject);
+        for (const key of pointKeys) {
+            if (!(key in targetObject) || isNaN(targetObject[key])) {
+                throw new Error('Object properties must be numeric.');
```
Aeri4a: Tip: You could also use .some() in line 282 and then if with throw error.

Aeri4a (approved):
Nice one 😄

Aeri4a (approved):

marcin-depcik (approved):

PR #888, opened 2025-02-05
[RPA-1905] Add email trigger whitelist interface and new logic

marcin-depcik:
## Description

image

image

image

## Types of changes

---


runbotics/runbotics-orchestrator-ui/src/src-app/views/tenants/TenantsListView/TenantsListEdit/AccordionSections/EmailTriggerSection/EmailTriggerWhitelistInput.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+        if (shouldAddWhitelistItem) handleAdd();
+    };
+
+    useEffect(() => {
+        if (input !== '' && !isInputValid(input)) {
+            setError(true);
+        } else {
+            setError(false);
+        }
```
bhszewczyk: I think you can shorted this to `setError(input !== '' && !isInputValid(input))`

runbotics/runbotics-orchestrator-ui/src/src-app/views/tenants/TenantsListView/TenantsListEdit/TenantsListEdit.utils.ts
```diff
@@ -1,19 +1,26 @@
...
 
+import { FormValidationState } from './TenantsListEdit.types';
+
 export const MINIMUM_NAME_CHARACTERS = 2;
 
 export const initialValidationState = {
     name: true,
     wasChanged: false,
+    wasTenantNameChanged: false,
+    wasWhitelistChanged: false,
```
bhszewczyk, line 10: not sure - why do we need that? maybe we can store this in wasChanged?

bhszewczyk: to support that, these 2 values are not being checked in checkFormFieldsValidation

bhszewczyk: ok, I see it is for payload mapping. so maybe wasChanged should be removed?

runbotics/runbotics-common/src/utils/emailTriggerWhitelistItemPattern.ts
```diff
@@ -0,0 +1,2 @@
+export const EMAIL_TRIGGER_WHITELIST_PATTERN =
+    /^(?:[a-zA-Z0-9.!#$%&*+/=?^_{|}~-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}|[a-zA-Z0-9.-]+\.[a-zA-Z]{2,})$/;
```
bhszewczyk: I would split this to 2 patterns: EMAIL_PATTERN and DOMAIN_PATTERN and then here I would combine these. It would be clear that we accept both without digging into regex :D

bhszewczyk (approved):

marcin-depcik (commented):
self CR

PR #899, opened 2025-02-12
[RPA-1960] Fixes issue with filtering in blog

bhszewczyk:
## Description

Filtering now works fine:
image

## Types of changes

---


star-in-the-jar (approved):

runbotics/runbotics-orchestrator-ui/src/contentful/common/utils.ts
```diff
@@ -132,3 +135,9 @@ export const getPaginatedUrl = (page: number, initialParams?: string): string =>
 
 export const checkIsDraft = (status: ContentfulContentPublishStatus): boolean => !status.publishedAt;
 
+const checkBlogPostMandatoryFields = (post: BlogPost) =>
```
Erquer, line 138: nice function, question is, if we want to know what fields are missing. if yes, you can do something like this function
`
private checkObject(object: TypedObject,) {
  const optionalFields = ['jobCategory', 'recruiterId'];
   const missingFields: string[] = [];
   for (const value of Object.entries(candidateObject)) {
  if (!value[1] && !optionalFields.includes(value[0])) {
  missingFields.push(value[0]);
   }
  }
   return missingFields;
   }
`

Erquer (approved):

runbotics/runbotics-orchestrator-ui/src/contentful/common/utils.ts
Erquer: second question, are these all the fields in BlogPost? if so, maybe consider using something like this
`
function hasAllFields<T extends object>(obj: Partial<T>, keys: (keyof T)[]): obj is T {
  return keys.every(key => key in obj);
}
`

PR #905, opened 2025-02-20
[RPA-1958] Add support for runbotics-plugins

marcin-depcik:
## Types of changes

---


runbotics/runbotics-desktop/.env.development
```diff
@@ -10,6 +10,7 @@ CFG_GECKO_DRIVER="./driver/windows/geckodriver-v0.29.1-win64"
 
 RUNBOTICS_LOGGER_FILE_MODE='true'
 RUNBOTICS_LOGGER_LEVEL='info'
-RUNBOTICS_EXTENSION_DIR='./extensions'
+RUNBOTICS_EXTENSION_DIR=''
```
Aeri4a: Is it okay to leave it without any directory?

runbotics/runbotics-desktop/src/core/bpm/desktop-runner/desktop-runner.service.ts
```diff
@@ -123,35 +155,35 @@ export class DesktopRunnerService implements OnModuleInit {
...
-                this.logger.log(`Success: Extension ${extension} loaded`);
+            this.logger.log('Number of external modules found: ' + externalModules.length);
+
+            for (const externalModule of externalModules) {
+                currentExternalModuleName = externalModule;
+                const externalModulePath = path.resolve(dirPath, externalModule);
+                this.logger.log(`Loading ${externalModule} external module`);
+                await moduleLoader.call(this, externalModulePath);
+                this.logger.log(`Success: External module ${externalModule} loaded`);
             }
```
Aeri4a: Should this loop be in try catch or try catch inside loop? What will happen if we have one invalid module and the other would be valid? What will be loaded?

runbotics/runbotics-desktop/src/core/bpm/desktop-runner/desktop-runner.service.ts
```diff
@@ -190,7 +222,66 @@ export class DesktopRunnerService implements OnModuleInit {
...
+            }
+
+            Object.defineProperty(internalHandler, PLUGIN_SERVICE, {
+                value: new handler(),
+                writable: true,
+            });
+
+            this.pluginHandlersMap.set(key, handler);
+            this.internalHandlersMap.set(internalHandlerKey, internalHandler);
+
```
Aeri4a, line 285: Why do we need to set internalHandlersMap to variables we previously got?

runbotics/runbotics-desktop/src/core/bpm/desktop-runner/desktop-runner.service.ts
```diff
@@ -280,4 +407,18 @@ export class DesktopRunnerService implements OnModuleInit {
...
+    private filterPlugins(dirPath: string, directoryEntry: Dirent) {
+        const isValidPluginsDirectory = this.detectPlugins(directoryEntry, dirPath);
+        if (!isValidPluginsDirectory) {
+            this.logger.warn(`Path "${dirPath}/${directoryEntry.name}" is not a valid plugins directory`);
+        }
+        return isValidPluginsDirectory;
+    }
+
+    private detectPlugins(directoryEntry: Dirent, extensionsPath: string) {
```
Aeri4a, line 425: What is type "Dirent"?

runbotics/runbotics-desktop/src/core/bpm/desktop-runner/desktop-runner.service.ts
```diff
@@ -112,7 +132,19 @@ export class DesktopRunnerService implements OnModuleInit {
         //     this.logger.warn('Hot reload is on! Remember to turn it off in production env.');
         // }
 
-        await this.loadExtensionsDirModules();
+        await this.loadExternalModules(
+            this.serverConfigService.pluginsDirPath,
+            this.filterPlugins,
+            this.loadPlugins,
+            'plugin actions'
```
Erquer: maybe set this strings as enum, or define values in functiona as union of strings so parameter has to have exact string value

runbotics/runbotics-desktop/src/core/bpm/desktop-runner/desktop-runner.service.ts
```diff
@@ -190,7 +222,66 @@ export class DesktopRunnerService implements OnModuleInit {
...
+        if (moduleEntries.length === 0) {
+            this.logger.error(`Module ${externalModule} handlers map has invalid structure or object is empty`);
+            return;
+        }
+
+        for (const [key, handler] of moduleEntries) {
+            if (!isPluginHandler(handler.prototype)) continue;
+
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore
```
Erquer, line 249: why ts-ignore?

runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Plugin/Plugin.thunks.ts
```diff
@@ -0,0 +1,41 @@
...
+                        )
+                    );
+
+                    return pluginModule.default;
+                })
+            )) as LoadedPlugin[];
+
+            return loadedPlugins;
+        } catch (error) {
+            if (!error.response) {
```
Erquer: is this check if this is AxiosError? if yes, you can check it as if( error instanceOf AxiosError

Erquer (approved):

runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Plugin/Plugin.thunks.ts
```diff
@@ -0,0 +1,41 @@
...
+                            new Blob([loadedPlugin.data], {
+                                type: 'application/javascript',
+                            })
+                        )
+                    );
+
+                    return pluginModule.default;
+                })
+            )) as LoadedPlugin[];
+
```
Aeri4a: If I am not wrong, you could use generic Promise.all<LoadedPlugin>(...)

Aeri4a (approved):

runbotics/runbotics-desktop/src/core/bpm/desktop-runner/desktop-runner.service.ts
marcin-depcik: because ‘internalHandlersMap’ accepts specific values as a key, and in this case we want to check whether some key of type string exists in the map

runbotics/runbotics-desktop/src/core/bpm/desktop-runner/desktop-runner.service.ts
marcin-depcik: in line `273` we are creating new property to inject new instance of loaded plugin handler. That's why we have to override internalHandlersMap

runbotics/runbotics-desktop/src/core/bpm/desktop-runner/desktop-runner.service.ts
marcin-depcik: type "Dirent" is response type of `readdirSync` function call

runbotics/runbotics-desktop/src/core/bpm/desktop-runner/desktop-runner.service.ts
marcin-depcik: good point

PR #924, opened 2025-03-11
[RPA-marketplace-paging-fix] MArketplace main page offer list fix

Erquer:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-orchestrator-ui/src/pages/marketplace/index.tsx
+++ b/runbotics/runbotics-orchestrator-ui/src/pages/marketplace/index.tsx
@@ -89,12 +89,13 @@ export const getServerSideProps: GetServerSideProps<Props> = async ({ query, loc
     }
     
     const totalPages = Math.ceil((offers?.length ?? 0) / DEFAULT_PAGE_SIZE);
+    const currentPageOffers = offers?.slice(0, DEFAULT_PAGE_SIZE) ?? [];
     
     return {
         props: {
             ...cache,
             metadata,
-            offers,
+            offers: currentPageOffers,
             page: {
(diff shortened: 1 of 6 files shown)


marcin-depcik (approved):

runbotics/runbotics-orchestrator-ui/src/pages/marketplace/index.tsx
```diff
@@ -89,12 +89,13 @@ export const getServerSideProps: GetServerSideProps<Props> = async ({ query, loc
     }
     
     const totalPages = Math.ceil((offers?.length ?? 0) / DEFAULT_PAGE_SIZE);
+    const currentPageOffers = offers?.slice(0, DEFAULT_PAGE_SIZE) ?? [];
```
star-in-the-jar, line 92: this way you'll always have DEFAULT_PAGE_SIZE amount of offers -> use pageSize ?? DEFAULT_PAGE_SIZE

star-in-the-jar: if there's an option to select the amount of elements on the page

star-in-the-jar (changes requested):

runbotics/runbotics-orchestrator-ui/src/pages/marketplace/index.tsx
Erquer: no there is not changable page size.  It is just a variable that is used in blog too. So I just reused it here because size of cards are similar so the number of elements are the same.

star-in-the-jar (approved):

PR #944, opened 2025-03-20
[RPA-2023] Incorrect name of the attached file downloaded on a Linux machine.

Hann4H:
## Types of changes

---

### DIFF
--- a/runbotics-orchestrator/build.gradle
+++ b/runbotics-orchestrator/build.gradle
@@ -26,7 +26,7 @@ plugins {
 
 group = "com.runbotics"
 
-version = "3.1.0-SNAPSHOT.50"
+version = "3.1.0-SNAPSHOT.51"
 description = ""
 
 sourceCompatibility=11
--- a/runbotics/common/config/runbotics.json
+++ b/runbotics/common/config/runbotics.json
@@ -1,3 +1,3 @@
 {
-    "version": "3.1.0-SNAPSHOT.50"
+    "version": "3.1.0-SNAPSHOT.51"
 }
--- a/runbotics/runbotics-desktop/package.json
+++ b/runbotics/runbotics-desktop/package.json
@@ -1,6 +1,6 @@
 {
     "name": "runbotics-desktop",
-    "version": "3.1.0-SNAPSHOT.50",
+    "version": "3.1.0-SNAPSHOT.51",
     "author": {
         "name": "runbotics",
         "email": "contact@runbotics.com"
--- a/runbotics/runbotics-desktop/src/action/mail/mail.action-handler.ts
+++ b/runbotics/runbotics-desktop/src/action/mail/mail.action-handler.ts
@@ -40,9 +40,7 @@ export default class MailActionHandler extends StatelessActionHandler {
 
         const attachment = input.attachment;
         if (attachment) {
-            const ext = path.extname(attachment);
-            const baseName = path.basename(attachment, ext);
-            const fileName = `${baseName}${ext}`;
+            const fileName = path.basename(attachment);
             await this.mailService.sendMail(
                 {
                     to: input.to,
--- a/runbotics/runbotics-desktop/src/action/mail/mail.action-handler.ts
+++ b/runbotics/runbotics-desktop/src/action/mail/mail.action-handler.ts
@@ -5,6 +5,7 @@ import { ServerConfigService } from '#config';
 import { MailCredential } from './mail.types';
 import fs from 'fs';
 import { credentialAttributesMapper } from '#utils/credentialAttributesMapper';
+import path from 'path';
 
 export type MailActionRequest =
     | DesktopRunRequest<'mail.send', MailSendActionInput>;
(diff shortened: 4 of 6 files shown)


runbotics/runbotics-desktop/src/action/file/file.action-handler.ts
```diff
@@ -61,11 +63,24 @@ export default class FileActionHandler extends StatelessActionHandler {
...
-            fs.closeSync(fs.openSync(`${input.path}`, 'a'));
-        } catch (err) {
-            throw Error(err);
-        }
+        if(fs.existsSync(input.path)) {
+            switch (input.conflict) {
+                case 'Overwrite': 
+                    await createNewFile(input.path);
+                    return 'File overwritten successfully';
+                case 'Extend name': {
```
star-in-the-jar: is the parnethesis necessary here?

runbotics/runbotics-desktop/src/action/file/file.action-handler.ts
```diff
@@ -61,11 +63,24 @@ export default class FileActionHandler extends StatelessActionHandler {
...
+                    const newFilePath = getUniqueFileName(input.path);
+                    await createNewFile(newFilePath);
+                    return 'File with extended name created successfully';
+                }
+                case 'Throw Error':
```
star-in-the-jar: I'd be better to have error code or at least an enum for error message returned by "fs.existsSync()" - misspelling can break the code

runbotics/runbotics-desktop/src/action/file/file.action-handler.ts
```diff
@@ -61,11 +63,24 @@ export default class FileActionHandler extends StatelessActionHandler {
...
+                    throw Error('File with the same name already exists');
+                default: 
+                    await createNewFile(input.path);
```
star-in-the-jar: leaving the success as default case can lead to hard debugging - there won't be error message and the log will say everything went successful, but the thing we want won't happen

runbotics/runbotics-desktop/src/action/file/file.action-handler.ts
```diff
@@ -114,3 +129,28 @@ export default class FileActionHandler extends StatelessActionHandler {
...
+
+function getUniqueFileName(filePath) {
+    const dir = path.dirname(filePath);
+    const ext = path.extname(filePath);
+    const baseName = path.basename(filePath, ext).replace(/\s\(\d+\)/, '');
+    
+    let counter = 1;
+    let newFilePath = path.join(dir, `${baseName} (${counter})${ext}`);
+    
+    // Keep increasing counter until we find a free name
```
star-in-the-jar: the comment is unncessary

runbotics/runbotics-desktop/src/action/file/file.action-handler.ts
```diff
@@ -114,3 +129,28 @@ export default class FileActionHandler extends StatelessActionHandler {
...
+        throw Error(err);
+    }
+}
+
+function getUniqueFileName(filePath) {
+    const dir = path.dirname(filePath);
+    const ext = path.extname(filePath);
+    const baseName = path.basename(filePath, ext).replace(/\s\(\d+\)/, '');
+    
+    let counter = 1;
```
star-in-the-jar: avoid modifiable variables (so like let instead const constructors)
do not use while loop - too risky

star-in-the-jar: imo instead of counting from 1 to infinity and checking file name each time - 1. get all file names 2. get the numbers from filenames with e.g. regex or splitting by '(' and getting last element  3. find the max number 4. create new file with a number that is one number higher

runbotics/runbotics-orchestrator-ui/src/src-app/Actions/file.actions.ts
```diff
@@ -90,6 +90,11 @@ const getFileActions = (): Record<string, IBpmnAction> => ({
                             //     'title': '(example.txt)',
                             //     'type': 'string'
                             // },
+                            conflict: {
+                                type: 'string',
+                                title: translate('Process.Details.Modeler.Actions.File.CreateFile.Conflict'),
+                                enum: ['Overwrite', 'Extend name', 'Throw Error'],
```
star-in-the-jar: because it's reused this should be typescript enum too

runbotics/runbotics-orchestrator-ui/src/src-app/Actions/file.actions.ts
```diff
@@ -101,10 +106,14 @@ const getFileActions = (): Record<string, IBpmnAction> => ({
...
                 'ui:order': ['input', 'output'],
+                conflict: {
+                    'ui:widget': 'radio',
+                },
             },
             formData: {
                 input: {
                     path: undefined,
+                    conflict: 'Overwrite'
                     // fileName: "",
```
star-in-the-jar: remove this oudated comment please

star-in-the-jar (changes requested):

runbotics/runbotics-desktop/src/action/file/file.action-handler.ts
Hann4H: yes

Erquer (approved):
@star-in-the-jar  I agree with his comment, so I approve PR, but please fix those comments before merging

star-in-the-jar (approved):
bump version

PR #947, opened 2025-03-21
[RPA-2014] Adding the 'Exists' action to the File section

Hann4H:
image

## Types of changes

---


runbotics/runbotics-desktop/src/action/file/file.action-handler.ts
```diff
@@ -97,6 +104,21 @@ export default class FileActionHandler extends StatelessActionHandler {
...
+
+        if (!input.name) {
+            throw new Error('Name is not provided.');
+        }
+
+        if (input.path && !fs.existsSync(input.path)) {
+            throw new Error('Provided path does not exist.');
+        }
+
+        const filePath = path.join(input.path, input.name);
```
star-in-the-jar, line 82: does it make sense to make path optional? is there a use case when the path is not required to complete action succesfully?

star-in-the-jar (approved):
when you're ready bump version

Erquer (approved):

PR #948, opened 2025-03-24
[RPA-1957] Allow tenant admin to manage credential collections

Aeri4a:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-orchestrator-ui/src/src-app/views/credentials/CredentialsCollection/EditCredentialsCollection/EditCredentialsCollection.utils.ts
+++ b/runbotics/runbotics-orchestrator-ui/src/src-app/views/credentials/CredentialsCollection/EditCredentialsCollection/EditCredentialsCollection.utils.ts
@@ -61,7 +61,7 @@ export const mapToEditCredentialCollectionDto = (collection: FrontCredentialColl
         accessType: collection.accessType,
         color: collection.color,
         sharedWith: sharedWithUsers,
-        description: collection.description ? collection.description : '',
+        description: collection.description ?? '',
         createdById: collection.createdById,
     };
 };
@@ -83,17 +83,28 @@ interface filterOptions {
 
 const SHARE_NOT_ALLOWED_ROLES = [Role.ROLE_GUEST, Role.ROLE_TENANT_ADMIN];
 
-export const filterSharableUsers = (value: string, allSharableUsers: UserDto[], {
-    sharedWithUsers =[],
-    selectedUsers = [],
-    collectionCreatorId,
-    currentUserId,
-}: filterOptions ): UserDto[] =>
+export const filterSharableUsers = (
+    value: string,
+    allSharableUsers: UserDto[],
+    {
+        sharedWithUsers = [],
+        selectedUsers = [],
+        collectionCreatorId,
+        currentUserId,
+    }: filterOptions
+): UserDto[] =>
     allSharableUsers.filter(
-        sharableUser =>
+        (sharableUser) =>
             sharableUser.email.toLowerCase().includes(value.toLowerCase()) &&
-        sharableUser.id !== collectionCreatorId && sharableUser.id !== currentUserId
-        && !sharableUser.roles.some(role => SHARE_NOT_ALLOWED_ROLES.includes(role)) &&
-        !sharedWithUsers?.some(sharedWithUser => sharedWithUser.email === sharableUser.email) &&
-        !selectedUsers?.some(selectedUser => selectedUser.email === sharableUser.email)
+            sharableUser.id !== collectionCreatorId &&
+            sharableUser.id !== currentUserId &&
+            !sharableUser.roles.some((role) =>
+                SHARE_NOT_ALLOWED_ROLES.includes(role)
+            ) &&
+            !sharedWithUsers?.some(
+                (sharedWithUser) => sharedWithUser.email === sharableUser.email
+            ) &&
+            !selectedUsers?.some(
+                (selectedUser) => selectedUser.email === sharableUser.email
+            )
     );
--- a/runbotics/runbotics-orchestrator-ui/src/src-app/views/credentials/CredentialsCollection/EditCredentialsCollection/EditCredentialsCollection.utils.ts
+++ b/runbotics/runbotics-orchestrator-ui/src/src-app/views/credentials/CredentialsCollection/EditCredentialsCollection/EditCredentialsCollection.utils.ts
@@ -1,9 +1,9 @@
-import { AccessType, DEFAULT_COLLECTION_COLOR, FrontCredentialCollectionDto, UserDto } from 'runbotics-common';
+import { AccessType, DEFAULT_COLLECTION_COLOR, FrontCredentialCollectionDto, Role, UserDto } from 'runbotics-common';
 
 import { translate } from '#src-app/hooks/useTranslations';
 
 import { SharedWithUser } from './SharedWithUsers/SharedWithUsers';
-import { EditCredentialsCollectionDto } from '../CredentialsCollection.types';
+import { EditCredentialsCollectionDto, EditCredentialsCollectionWithCreatorDto } from '../CredentialsCollection.types';
 
 export enum InputErrorType {
     NAME_IS_REQUIRED = 'NAME_IS_REQUIRED'
@@ -23,14 +23,15 @@ export const initialFormValidationState: CollectionFormValidation = {
     name: false
 };
 
-export const initialCredentialsCollectionData: EditCredentialsCollectionDto = {
+export const initialCredentialsCollectionData: EditCredentialsCollectionWithCreatorDto = {
     name: '',
     accessType: AccessType.PRIVATE,
     description: '',
-    color: DEFAULT_COLLECTION_COLOR
+    color: DEFAULT_COLLECTION_COLOR,
+    createdById: null,
 };
 
-export const getInitialCredentialsCollectionData = (collection: null | EditCredentialsCollectionDto): EditCredentialsCollectionDto => {
+export const getInitialCredentialsCollectionData = (collection: null | EditCredentialsCollectionWithCreatorDto): EditCredentialsCollectionWithCreatorDto => {
     if (!collection) {
         return initialCredentialsCollectionData;
     }
@@ -40,11 +41,12 @@ export const getInitialCredentialsCollectionData = (collection: null | EditCrede
         accessType: collection.accessType,
         color: collection.color,
         sharedWith: collection.sharedWith,
-        description: collection.description
+        description: collection.description,
+        createdById: collection.createdById,
     };
 };
 
-export const mapToEditCredentialCollectionDto = (collection: FrontCredentialCollectionDto): EditCredentialsCollectionDto => {
+export const mapToEditCredentialCollectionDto = (collection: FrontCredentialCollectionDto): EditCredentialsCollectionWithCreatorDto => {
     const sharedWithUsers = collection.credentialCollectionUser
         ? [...collection.credentialCollectionUser]
             .filter(credentialUser => credentialUser.userId !== collection.createdById)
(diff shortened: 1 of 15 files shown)


runbotics/runbotics-orchestrator-ui/src/src-app/views/credentials/CredentialsCollection/EditCredentialsCollection/EditCredentialsCollection.utils.ts
```diff
@@ -23,14 +23,15 @@ export const initialFormValidationState: CollectionFormValidation = {
...
 };
 
-export const initialCredentialsCollectionData: EditCredentialsCollectionDto = {
+export const initialCredentialsCollectionData: EditCredentialsCollectionWithCreatorDto = {
     name: '',
     accessType: AccessType.PRIVATE,
     description: '',
-    color: DEFAULT_COLLECTION_COLOR
+    color: DEFAULT_COLLECTION_COLOR,
+    createdById: null,
```
Erquer, line 31: type mismatch  - you defined type number on interface, but here is null. Change the interface field to number | null maybe, to avoid errors

runbotics/runbotics-orchestrator-ui/src/src-app/views/credentials/CredentialsCollection/EditCredentialsCollection/EditCredentialsCollection.utils.ts
```diff
@@ -40,11 +41,12 @@ export const getInitialCredentialsCollectionData = (collection: null | EditCrede
         accessType: collection.accessType,
         color: collection.color,
         sharedWith: collection.sharedWith,
-        description: collection.description
+        description: collection.description,
```
Erquer, line 44: why doing this destructurization in return statement if collection is type of returned object. return collection is enough imo.

runbotics/runbotics-orchestrator-ui/src/src-app/views/credentials/CredentialsCollection/EditCredentialsCollection/EditCredentialsCollection.utils.ts
```diff
@@ -59,11 +61,12 @@ export const mapToEditCredentialCollectionDto = (collection: FrontCredentialColl
         accessType: collection.accessType,
         color: collection.color,
         sharedWith: sharedWithUsers,
-        description: collection.description ? collection.description : ''
+        description: collection.description ? collection.description : '',
```
Erquer: collection.description ?? ''

runbotics/runbotics-orchestrator-ui/src/src-app/views/credentials/CredentialsCollection/EditCredentialsCollection/GeneralOptions/GeneralOptions.tsx
```diff
@@ -15,8 +15,8 @@ interface CredentalsCollectionGeneralOptionsProps {
     setFormValidationState: (formValidationState: CollectionFormValidation) => void;
     inputErrorType: InputErrorType;
     setInputErrorType: (errorType: InputErrorType) => void;
-    formState: EditCredentialsCollectionDto;
-    setFormState: (state: ((prevState: EditCredentialsCollectionDto) => EditCredentialsCollectionDto)) => void;
+    formState: EditCredentialsCollectionWithCreatorDto;
+    setFormState: (state: ((prevState: EditCredentialsCollectionWithCreatorDto) => EditCredentialsCollectionWithCreatorDto)) => void;
```
Erquer: this function is Dispatch<SetStateAction<EditCredentialsCollectionWithCreatorDto[]>> type. it's original setState action type if I understood this function type correctry

runbotics/runbotics-orchestrator-ui/src/src-app/views/credentials/CredentialsCollection/EditCredentialsCollection/SharedWithUsers/SharedWithUsers.tsx
```diff
@@ -21,8 +21,8 @@ export interface SharedWithUser {
 }
 
 interface SharedWithUsersProps {
-    credentialsCollectionFormState: EditCredentialsCollectionDto;
-    setCredentialsCollectionFormState: (state: (prevState: EditCredentialsCollectionDto) => EditCredentialsCollectionDto) => void;
+    credentialsCollectionFormState: EditCredentialsCollectionWithCreatorDto;
+    setCredentialsCollectionFormState: (state: (prevState: EditCredentialsCollectionWithCreatorDto) => EditCredentialsCollectionWithCreatorDto) => void;
```
Erquer: as above

runbotics/runbotics-scheduler/src/scheduler-database/credential-collection/credential-collection.service.ts
```diff
@@ -210,19 +239,29 @@ export class CredentialCollectionService {
...
+                await Promise.all(
+                    credentialCollectionUserArrayToSave.map((ccu) =>
+                        manager.save(CredentialCollectionUser, ccu)
+                    )
+                );
+
+                await manager.update(CredentialCollection, id, dto).catch(async error => {
+                    await this.throwErrorIfNameTaken(user.tenantId, dto.name, user.id, { id });
+
+                    throw new BadRequestException(error.message);
```
Erquer: if you throw error here, did you check if transaction is fully reverted?

runbotics/runbotics-scheduler/src/scheduler-database/credential-collection/credential-collection.service.ts
```diff
@@ -261,40 +302,48 @@ export class CredentialCollectionService {
     private async getCollectionUserArrayWithPrivileges(
         sharedWith: UpdateCredentialCollectionDto['sharedWith'],
         credentialCollection: CredentialCollection,
-        user: Omit<User, 'authorities'>,
         tenantId: Tenant['id'],
     ) {
-        const credentialCollectionCreator = this.credentialCollectionUserRepository.create({
```
Erquer, line 267: I don't exacly know the entity of credentialCollectionUser. But if the credentialCollectionId is the primary key - check if this works fine. For me there could be an error thrown if this field is primary key. so maybe better solution would be use check if there is creator and then skip the creation or use save function to update the particular creator

runbotics/runbotics-scheduler/src/scheduler-database/credential-collection/credential-collection.service.ts
```diff
@@ -261,40 +302,48 @@ export class CredentialCollectionService {
...
+            throw new BadRequestException('Some emails are not correct');
+        }
 
-        const unknownEmails = userEmails.filter(email => !grantAccessEmails.has(email));
+        const grantAccessUsers = await this.userService.findByEmailsNotTenantAdmin(userEmails, tenantId);
 
-        if (unknownEmails.length > 0) {
-            throw new BadRequestException(`Users with emails ${unknownEmails.join(', ')} do not exist`);
-        }
+        return grantAccessUsers.map(user => {
```
Erquer, line 330: why map? in this case it sshould be forEach imo. 
Or the best case 
repository.save(grantAccessUsers.map(user => return object to save based on conditions).
The save function saves all the given users in 1 transaction. In your solution there is new transaction for every insert

Erquer (changes requested):

runbotics/runbotics-orchestrator-ui/src/src-app/views/credentials/CredentialsCollection/EditCredentialsCollection/EditCredentialsCollection.utils.ts
bhszewczyk: we do not use strict so it should not result in error, but still I concur it will be a good indicator for devs that it could be nullable (and I think a good practice for us all)

runbotics/runbotics-orchestrator-ui/src/src-app/views/credentials/CredentialsCollection/EditCredentialsCollection/EditCredentialsCollection.utils.ts
```diff
@@ -75,18 +78,22 @@ interface filterOptions {
...
+    collectionCreatorId,
+    currentUserId,
 }: filterOptions ): UserDto[] =>
     allSharableUsers.filter(
         sharableUser =>
             sharableUser.email.toLowerCase().includes(value.toLowerCase()) &&
-        sharableUser.id !== collectionCreatorId
-        &&
+        sharableUser.id !== collectionCreatorId && sharableUser.id !== currentUserId
+        && !sharableUser.roles.some(role => SHARE_NOT_ALLOWED_ROLES.includes(role)) &&
```
bhszewczyk: consider moving && to another line to keep previous formatting

star-in-the-jar (approved):

bhszewczyk (changes requested):

runbotics/runbotics-orchestrator-ui/src/src-app/views/credentials/CredentialsCollection/EditCredentialsCollection/EditCredentialsCollection.utils.ts
Aeri4a: I am not sure now actually. Because the only difference is that in this code we skip 'id' if exists in object

runbotics/runbotics-orchestrator-ui/src/src-app/views/credentials/CredentialsCollection/EditCredentialsCollection/GeneralOptions/GeneralOptions.tsx
Aeri4a: You're right. I will complete those types.

runbotics/runbotics-scheduler/src/scheduler-database/credential-collection/credential-collection.service.ts
Aeri4a: It should be correct, because:
1. If update function throws error it will be catched in '.catch()'
2. In catch function either will be thrown error from calling function in first line or BadRequestException.
3. Those two errors won't be catched by update function nor .catch() so they will propagate to transaction scope
4. In TypeORM library - transaction body will catch this error and make a revert of whole transaction

runbotics/runbotics-scheduler/src/scheduler-database/credential-collection/credential-collection.service.ts
Aeri4a: collection_id and user_id both stand as primary key.

In this function we get all user entities group to share (including creator and excluding any tenant admin).
After that in create or update function, all previous sharable records will be erased for that collection and replaced with that new group.
As continuation for previous code logic, I don't intersect previous users with new ones and save/delete those who differ but replace all previous with all new.

runbotics/runbotics-scheduler/src/scheduler-database/credential-collection/credential-collection.service.ts
Aeri4a: That is a good point for save function. I think I can use it in update/create function. In case of this function i am doing map because I need to convert user objects to user entity (repo.create() is not inserting it to database). So the whole logic of saving to db is performed in service create/update function.

runbotics/runbotics-orchestrator-ui/src/src-app/views/credentials/CredentialsCollection/EditCredentialsCollection/EditCredentialsCollection.utils.ts
Aeri4a: Good, I formatted it now with prettier automatically

Erquer (approved):

bhszewczyk (approved):

PR #963, opened 2025-04-11
[RPA-2018] The action 'File: Create' should not overwrite an existing file.

teawithsand:
## Description

Overwrite now overrides file contents. Before it would leave it unmodified.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

---

### DIFF
--- a/runbotics/runbotics-desktop/src/action/file/file.utils.ts
+++ b/runbotics/runbotics-desktop/src/action/file/file.utils.ts
@@ -2,9 +2,9 @@ import fs from 'fs';
 import path from 'path';
 import { ActionRegex } from 'runbotics-common';
 
-export async function createNewFile(path: string){
+export async function createNewFile(path: string, truncate = false) {
     try {
-        fs.closeSync(fs.openSync(`${path}`, 'a'));
+        fs.closeSync(fs.openSync(`${path}`, truncate ? 'w' : 'a'));
     } catch (err) {
         throw Error(err);
     }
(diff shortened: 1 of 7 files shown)


Erquer (approved):

runbotics/runbotics-desktop/src/action/file/file.utils.ts
```diff
@@ -2,9 +2,9 @@ import fs from 'fs';
 import path from 'path';
 import { ActionRegex } from 'runbotics-common';
 
-export async function createNewFile(path: string){
+export async function createNewFile(path: string, truncate = false) {
```
bhszewczyk, line 5: I would think of renaming truncate to something crystal clear like `overwrite` or `replaceContent`

bhszewczyk (changes requested):
I know it wasn't exactly included in the task, but could you please add tooltip for the user to explain what "Conflict" means?

runbotics/runbotics-desktop/src/action/file/file.utils.ts
teawithsand: Truncate is typical word using in this kind of context judging from my experience. For instance Node.js documentation explains 'w' flag as:
`'w': Open file for writing. The file is created (if it does not exist) or truncated (if it exists).`

Also this function only creates a file, without returning any kind of description or otherwise providing a method to put some kind of contents into this file, which is why I believe that if parameter name change is required, a name like `discardExistingContent` or `ensureEmpty` would be more fitting in this case.

bhszewczyk: I get it, but from just reading "truncate" it suggest it is, well... truncated (so cut off) not re-written. I like `discardExistingContent` name

bhszewczyk (approved):

PR #979, opened 2025-04-29
[RPA-2003] Adding a notification about account activation (updated)

teawithsand:
Added default (preselected) reason of user decline. 

This PR also includes all changes of https://github.com/runbotics/runbotics/pull/950

image

## Types of changes

---

### DIFF
--- a/runbotics/runbotics-common/src/model/api/user.model.ts
+++ b/runbotics/runbotics-common/src/model/api/user.model.ts
@@ -21,7 +21,6 @@ export interface User {
     lastModifiedBy: string;
     tenantId: string;
     authorities: IAuthority[];
-    message?: string;
 }
 
 export type BasicUserDto = Pick<User, 'id' | 'email'>;
@@ -41,3 +40,10 @@ export type UserDto = Omit<
     roles: Role[];
     featureKeys: FeatureKey[];
 };
+
+export type ActivateUserDto = {
+    id: number
+    roles: Role[],
+    message?: string
+    tenantId: string
+}
--- a/runbotics/runbotics-common/src/model/api/user.model.ts
+++ b/runbotics/runbotics-common/src/model/api/user.model.ts
@@ -21,6 +21,7 @@ export interface User {
     lastModifiedBy: string;
     tenantId: string;
     authorities: IAuthority[];
+    message?: string;
 }
 
 export type BasicUserDto = Pick<User, 'id' | 'email'>;
--- a/runbotics/runbotics-orchestrator-ui/src/src-app/utils/ApiTenantResource.ts
+++ b/runbotics/runbotics-orchestrator-ui/src/src-app/utils/ApiTenantResource.ts
@@ -11,7 +11,6 @@ export interface PayloadWrap<T> {
     payload?: T;
     resourceId?: string | number;
     pageParams?: PageRequestParams;
-    message?: string;
 }
 
 type ResourcePath = string | ((resourceId: string | number) => string);
--- a/runbotics/runbotics-orchestrator-ui/src/src-app/utils/ApiTenantResource.ts
+++ b/runbotics/runbotics-orchestrator-ui/src/src-app/utils/ApiTenantResource.ts
@@ -11,6 +11,7 @@ export interface PayloadWrap<T> {
     payload?: T;
     resourceId?: string | number;
     pageParams?: PageRequestParams;
+    message?: string;
 }
 
 type ResourcePath = string | ((resourceId: string | number) => string);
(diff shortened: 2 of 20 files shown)


runbotics/runbotics-orchestrator-ui/src/src-app/utils/ApiTenantResource.ts
```diff
@@ -11,6 +11,7 @@ export interface PayloadWrap<T> {
     payload?: T;
     resourceId?: string | number;
     pageParams?: PageRequestParams;
+    message?: string;
 }
```
Aeri4a: I don't like the fact that this general interface contains additional message. Why it is not in payload?

runbotics/runbotics-common/src/model/api/user.model.ts
```diff
@@ -21,6 +21,7 @@ export interface User {
     lastModifiedBy: string;
     tenantId: string;
     authorities: IAuthority[];
+    message?: string;
```
Aeri4a: Why is message in a user model?

runbotics/runbotics-scheduler/src/scheduler-database/user/user.service.ts
```diff
@@ -166,6 +166,8 @@ export class UserService {
             lastModifiedBy: executor.email,
         };
 
+        this.mailService.sendUserAcceptMail(user, userDto.message);
+
```
Aeri4a: There is a lack of IF for checking isFirstActivation, because we would like to be notified only the FIRST time after activation

runbotics/runbotics-common/src/model/api/user.model.ts
Aeri4a: If you need to have one more field for updating, just use another dto like those below (BasicUserDto)

runbotics/runbotics-orchestrator-ui/src/src-app/views/users/DeclineUserReasonSelector/DeclineUserReasonSelector.tsx
```diff
@@ -80,6 +87,10 @@ export const DeclineUserReasonSelector = ({
                 )}
             </Typography>
             <StyledRadioGroup value={reason} onChange={handleRadioControlChange}>
+                <RadioControl
+                    label={declineReason[DeclineReason.DEFAULT_REASON].message}
+                    value={declineReason[DeclineReason.DEFAULT_REASON].value}
+                />
                 <RadioControl
                     label={declineReason[DeclineReason.FIRST_REASON].message}
```
Aeri4a: In my opinion user don't have to know the default message

runbotics/runbotics-scheduler/src/scheduler-database/user/dto/update-user.dto.ts
```diff
@@ -8,6 +8,11 @@ export const updateUserSchema = z.object({
...
     tenantId: z.string(),
     roles: z.array(z.nativeEnum(Role)),
+    message: z.string().trim().min(1).optional(),
 }).partial();
 
 export type UpdateUserDto = z.infer<typeof updateUserSchema>;
+
+export const message =  z.string().trim().min(1).optional();
+
+export type Message = z.infer<typeof message>;
```
Aeri4a: Include message in update user dto as optional and actually I would suggest to change the name from message to for example "decline reason" or "activation message" for clean code understanding.

runbotics/runbotics-scheduler/src/mail/mail.service.ts
```diff
@@ -177,6 +178,16 @@ export class MailService {
         }
     }
```
Aeri4a, line 178: In this file in line 171 you should also delete checking field in dto. I think you should merge this function with function you called sendUserAcceptMail.

Aeri4a (changes requested):

runbotics/runbotics-scheduler/src/scheduler-database/user/user.controller.ts
```diff
@@ -49,6 +50,20 @@ export class UserController {
...
+        @Param('id', ParseIntPipe) userId: User['id'],
+        @Body(new ZodValidationPipe(activateUserSchema)) userDto: ActivateUserDto,
+    ) {
+        return this.userService.activate(
+            userDto,
+            userId,
+            user
+        );
+    }
+
```
Aeri4a, line 66: There is no handling for situation in which user dto comes with different tenant id than it is in url. That situation should not happen.

Aeri4a (changes requested):

Aeri4a (approved):
nice, eventually clean code

Aeri4a (approved):

igorspychalaafo (approved):

bhszewczyk (approved):

janeksmielowski (approved):

janeksmielowski:
why exactly do we have predefined statements for reject reason? I believe we should have one message: Your account activation has been rejected. With optional reason attached

PR #981, opened 2025-05-06
[RPA-2192] Properly catch error when process does not exist under user's tenant

bhszewczyk:
## Description

- properly catches error in case the process under tenant does not exist or process exists but under different tenant (in both cases there is 404)

## Types of changes

---

### DIFF
--- a/runbotics/runbotics-scheduler/src/scheduler-database/process/process.service.ts
+++ b/runbotics/runbotics-scheduler/src/scheduler-database/process/process.service.ts
@@ -35,7 +35,7 @@ export class ProcessService {
         return new ProcessService(
             em.getRepository(ProcessEntity),
             this.processCollectionService.withEntityManager(em)
-        )
+        );
     }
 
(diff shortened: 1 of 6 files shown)


runbotics/runbotics-scheduler/src/scheduler-database/process/process.service.ts
```diff
@@ -102,7 +102,14 @@ export class ProcessService {
...
 
     async hasAccess(user: User, processId: number): Promise<boolean> {
-        const process = await this.processRepository.findOneByOrFail({id: processId});
+        const process = await this.processRepository
+            .findOneByOrFail({
+                id: processId,
+                tenantId: user.tenantId
+            })
+            .catch(() => {
+                throw new NotFoundException();
```
Erquer: maybe Add some message to this error like No Process found, so the source of the error will be known

Erquer (approved):

Aeri4a (approved):

PR #1029, opened 2025-06-09
[RPA-2276] Fix admin can't activate users

teawithsand:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Users/Users.thunks.ts
+++ b/runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Users/Users.thunks.ts
@@ -48,7 +48,7 @@ export const update = createAsyncThunk<UserDto, PartialUserDto>(
 
 export const activate = createAsyncThunk<void, ActivateUserDto>(
     'users/activate',
-    (activateData) => axios.post<ActivateUserDto>(`/api/scheduler/users/${activateData.id}/activate`, activateData)
+    (activateData) => axios.post<ActivateUserDto>(`/api/scheduler/users/activate/${activateData.id}`, activateData)
         .then((response => void response))
 );
 
--- a/runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Users/Users.thunks.ts
+++ b/runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Users/Users.thunks.ts
@@ -48,7 +48,7 @@ export const update = createAsyncThunk<UserDto, PartialUserDto>(
 
 export const activate = createAsyncThunk<void, ActivateUserDto>(
     'users/activate',
-    (activateData) => axios.patch<ActivateUserDto>(`/api/scheduler/users/${activateData.id}/activation`, activateData)
+    (activateData) => axios.patch<ActivateUserDto>(`/api/scheduler/users/${activateData.id}/activate`, activateData)
         .then((response => void response))
 );
 
@@ -57,7 +57,7 @@ export const updateInTenant = ApiTenantResource.patch<UserDto, PartialUserDto>(
 );
 
 export const activateInTenant = ApiTenantResource.patch<UserDto, ActivateUserDto>(
-    'users/activateInTenant', (id) => `${USERS_PATH}/${id}/activation`
+    'users/activateInTenant', (id) => `${USERS_PATH}/${id}/activate`
 );
 
 export const deleteUser = createAsyncThunk<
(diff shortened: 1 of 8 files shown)


runbotics/runbotics-scheduler/src/scheduler-database/user/user.service.ts
```diff
@@ -287,16 +286,21 @@ export class UserService {
...
-        const TENANT_ALLOWED_ROLES = [
-            Role.ROLE_USER,
-            Role.ROLE_TENANT_ADMIN,
-            Role.ROLE_EXTERNAL_USER,
-        ];
-
-        if (!isTenantAdmin(user) && !TENANT_ALLOWED_ROLES.includes(roles[0])) {
-            throw new BadRequestException('Wrong role');
+
+        if (isAdmin(user)) {
```
bhszewczyk: as discussed during meetings for role permissions we preferably use feature keys instead of role. i think using here MANAGE_ALL_TENANTS && MANAGE_INACTIVE_USERS for admin and MANAGE_INACTIVE_USERS for tenant admin is reasonable here as you also use it in controller

runbotics/runbotics-scheduler/src/scheduler-database/user/user.controller.ts
```diff
@@ -96,7 +96,7 @@ export class UserController {
         return this.userService.update(userDto, userId, user);
     }
 
-    @Patch('users/activate/:id')
+    @Post('users/activate/:id')
```
bhszewczyk: tbh not sure why post rather that patch?

bhszewczyk (changes requested):

runbotics/runbotics-scheduler/src/scheduler-database/user/user.service.ts
```diff
@@ -287,16 +286,19 @@ export class UserService {
...
+
         const TENANT_ALLOWED_ROLES = [
             Role.ROLE_USER,
             Role.ROLE_TENANT_ADMIN,
             Role.ROLE_EXTERNAL_USER,
         ];
 
-        if (!isTenantAdmin(user) && !TENANT_ALLOWED_ROLES.includes(roles[0])) {
-            throw new BadRequestException('Wrong role');
+        if (!this.hasFeatureKey(user, FeatureKey.MANAGE_ALL_TENANTS)) {
```
teawithsand: User with MANAGE_ALL_TENANTS role can now grant any role, including ADMIN, which tenant-admin shouldn't be able to do, since they're limited to these three roles above. At least that's what UI would indicate.

runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Users/Users.thunks.ts
```diff
@@ -48,7 +48,7 @@ export const update = createAsyncThunk<UserDto, PartialUserDto>(
 
 export const activate = createAsyncThunk<void, ActivateUserDto>(
     'users/activate',
-    (activateData) => axios.post<ActivateUserDto>(`/api/scheduler/users/${activateData.id}/activate`, activateData)
+    (activateData) => axios.patch<ActivateUserDto>(`/api/scheduler/users/activate/${activateData.id}`, activateData)
```
janeksmielowski: patch seems more adequate here, but just to mention, in RESTful API the action/function/procedure name should not exist in the URL. We could have action query param. However, if we want to keep this approach, I would leave the action as last segment: `/api/scheduler/users/${activateData.id}/activate` seems better to me

Source:
image

janeksmielowski (approved):

Fiflon (approved):
Bump version :))

bhszewczyk (approved):
remember to bump version

runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Users/Users.thunks.ts
```diff
@@ -48,16 +48,16 @@ export const update = createAsyncThunk<UserDto, PartialUserDto>(
 
 export const activate = createAsyncThunk<void, ActivateUserDto>(
     'users/activate',
-    (activateData) => axios.post<ActivateUserDto>(`/api/scheduler/users/${activateData.id}/activate`, activateData)
+    (activateData) => axios.patch<ActivateUserDto>(`/api/scheduler/users/${activateData.id}/activation`, activateData)
```
janeksmielowski: should be a verb, like a name of function, so 'activation' is incorrect, because it's a noun

janeksmielowski (changes requested):

Aeri4a (approved):

janeksmielowski (approved):

PR #1032, opened 2025-06-09
[RPA-2189.2] Fixes issue where bot collection couldn't be updated without changing the collection name

bhszewczyk:
## Description

- previously there was an error during modifying collection without changing the name (e.g. user wanted to share collection with users)
- now: collection can be updated without changing the collection name
- adjusted UI to receive messages from backend (validation errors, collection edit errors)
- adjusted UI to filter out admin role from sharable users

## Types of changes

---


teawithsand (approved):

runbotics/runbotics-orchestrator-ui/src/src-app/views/bot/BotCollectionView/Dialog/modify/BotCollectionModifyDialog.tsx
```diff
@@ -77,7 +78,7 @@ const BotCollectionModifyDialog: FC<ModifyBotCollectionDialogProps> = ({ collect
         const body = createCollectionEntityToSend();
         const { type, payload } = await updateCollection(body);
 
-        if (type === REJECT_REQUEST_TYPE) {
+        if (type === REJECT_REQUEST_TYPE || type === REJECT_UPDATE_REQUEST_TYPE) {
```
janeksmielowski: not sure if it makes it cleaner approach, but worth a consideration for future enhancements instead joining with OR: `[REJECT_REQUEST_TYPE, REJECT_UPDATE_REQUEST_TYPE].includes(type)`

runbotics/runbotics-scheduler/src/scheduler-database/bot-collection/bot-collection.service.ts
```diff
@@ -75,13 +75,15 @@ export class BotCollectionService {
             throw new NotFoundException();
         });
 
+        if (collection.name !== collectionDto.name) {
```
janeksmielowski, line 78: maybe we should call it newCollection, because collection seems not enough to understand the difference between between collectionDto

janeksmielowski (changes requested):

runbotics/runbotics-scheduler/src/scheduler-database/bot-collection/bot-collection.service.ts
bhszewczyk: I think this is consistent with other services (e.g. process-crud.service).
newCollection (in create) and updatedCollection (in update) is assigned once the dto which came is actually saved (you can check out whole method)

janeksmielowski: Then I would say it's consistently incorrect in whole project :D As discussed on the call - let's skip it for now, and we'll get back to it when implementing process collections

janeksmielowski (approved):

PR #1043, opened 2025-06-18
[RPA-2308] Added unit testing environment for Orchestrator UI

teawithsand:
Provides example unit test, so that vitest won't complain that there are no tests.

In the future it can be removed.

## Types of changes

---


Erquer (approved):

runbotics/runbotics-orchestrator-ui/src/example.test.tsx
bhszewczyk, line 1: just a suggestion - if this is an example maybe it is worth to extract this to folder testingExample and there Counter extract to other file and tests leave here? My thoughts makes sense if we actually decided we will be doing tests inside the folder where the actual component is. If that changes during my absence - ignore my comment. :)

bhszewczyk (approved):

bhszewczyk (approved):

PR #1053, opened 2025-07-03
[RPA-2299] Fix history table and hasAccess function

Fiflon:
## Description

The bot's history was not working properly. During my research, I found out that the hasAccess function was not working properly either.
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-scheduler/src/scheduler-database/process-collection/process-collection.service.ts
+++ b/runbotics/runbotics-scheduler/src/scheduler-database/process-collection/process-collection.service.ts
@@ -84,7 +84,7 @@ export class ProcessCollectionService {
         return (hasAccessToParent && hasAccess);
     }
 
-    async hasAccess(user: User, collectionId: string): Promise<boolean> {
+    async hasAccess(user: User, collectionId?: string): Promise<boolean> {
         const hasAllAccess = hasFeatureKey(
             user,
             FeatureKey.PROCESS_COLLECTION_ALL_ACCESS
@@ -92,7 +92,7 @@ export class ProcessCollectionService {
 
         if (hasAllAccess) return true;
 
-        if (collectionId === null) {
+        if (!collectionId) {
             return true;
         }
 
@@ -115,7 +115,7 @@ export class ProcessCollectionService {
                 })
                 .catch(() => false)
                 .then((res) => {
-                    return res === false ? false : true;
+                    return !!res;
                 }))
         );
     }
--- a/runbotics/runbotics-scheduler/src/scheduler-database/process-collection/process-collection.service.ts
+++ b/runbotics/runbotics-scheduler/src/scheduler-database/process-collection/process-collection.service.ts
@@ -1,15 +1,12 @@
-import { BadGatewayException, BadRequestException, ForbiddenException, Injectable, NotFoundException } from '@nestjs/common';
+import { BadRequestException, ForbiddenException, Injectable, NotFoundException } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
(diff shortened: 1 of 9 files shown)


runbotics/runbotics-scheduler/src/scheduler-database/process-collection/process-collection.service.ts
```diff
@@ -69,38 +66,59 @@ export class ProcessCollectionService {
...
-        const hasAllAccess = hasFeatureKey(user, FeatureKey.PROCESS_COLLECTION_ALL_ACCESS);
+        const hasAllAccess = hasFeatureKey(
+            user,
+            FeatureKey.PROCESS_COLLECTION_ALL_ACCESS
+        );
+
+        if (hasAllAccess) return true;
 
-        if (hasAllAccess)
+        if (collectionId === null) {
```
Erquer: make parameter optional or type string | null, because now you teoreticly check non existing case. If you make optional param, better solution will be if(!collectionId) so the undefined value will be served also.

runbotics/runbotics-scheduler/src/scheduler-database/process-collection/process-collection.service.ts
```diff
@@ -69,38 +66,59 @@ export class ProcessCollectionService {
...
+                            id: collectionId,
+                        },
+                        {
+                            createdBy: Equal(user.id),
+                        },
+                    ],
+                })
+                .catch(() => false)
+                .then((res) => {
+                    return res === false ? false : true;
```
Erquer: you can simplify by doing return !!res, it will convert res to boolean

runbotics/runbotics-scheduler/src/scheduler-database/process-instance/process-instance.service.ts
```diff
@@ -5,44 +5,53 @@ import {
...
 type MappedProcessInstance = ProcessInstance & { hasSubprocesses: boolean };
 
 const RELATIONS: FindOptionsRelations<ProcessInstance> = {
     bot: true,
-    process: true,
+    process: {
+        processCollection: {
+            users: true,
+            createdBy: true,
+
```
Erquer: formating

runbotics/runbotics-scheduler/src/scheduler-database/process-instance/process-instance.service.ts
```diff
@@ -84,29 +93,85 @@ export class ProcessInstanceService {
...
+        };
+
+
         if (
             !isTenantAdmin(user) &&
-            (
-                specs.where.processId === undefined ||
-                !await this.processService.hasAccess(user, +processId)
-            )
+            specs.where.processId === undefined &&
```
Erquer: i would rather separate this conditions, because if you have undefined value it doesn't mean you are forbidden to see resource, you sent bad request. so imho it should be split to 2 different ifs 1 with ForbiddenException, second on with BadRequestException, or InternalServerException. The one with undefined values should be checked 1st

Erquer (changes requested):

runbotics/runbotics-scheduler/src/scheduler-database/process-instance/process-instance.service.ts
```diff
@@ -138,16 +136,23 @@ export class ProcessInstanceService {
             },
         };
 
+        if (
+            specs.where.processId === undefined &&
+            specs.where.botId === undefined
+        ) {
+            throw new BadRequestException(
+                'Either processId or botId must be provided'
```
Erquer, line 147: processId AND botId

Erquer (approved):

bhszewczyk (approved):

PR #1058, opened 2025-07-04
[RPA-2310] Add process statistics service

igorspychalaafo:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-scheduler/src/scheduler-database/process-instance-event/process-instance-event.service.ts
+++ b/runbotics/runbotics-scheduler/src/scheduler-database/process-instance-event/process-instance-event.service.ts
@@ -93,7 +93,9 @@ export class ProcessInstanceEventService {
     async findAllByProcessInstanceId(processInstanceId: ProcessInstance['id']) {
         return this.processInstanceEventRepository.find({
             where: {
-                processInstance: {id: processInstanceId}
+                processInstance: { 
+                    id: processInstanceId 
+                }
             },
             relations: RELATIONS,
         });
--- a/runbotics/runbotics-scheduler/src/scheduler-database/process-instance-event/process-instance-event.service.ts
+++ b/runbotics/runbotics-scheduler/src/scheduler-database/process-instance-event/process-instance-event.service.ts
@@ -90,6 +90,15 @@ export class ProcessInstanceEventService {
         });
     }
 
+    async findAllByProcessInstanceId(processInstanceId: ProcessInstance['id']) {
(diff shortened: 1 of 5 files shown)


Erquer (approved):

runbotics/runbotics-scheduler/src/utils/process-statistics/process-statistics.service.ts
```diff
@@ -0,0 +1,61 @@
...
+                    const duration = new Date(event.finished).getTime() - new Date(event.created).getTime();
+                    totalDuration += duration;
+                    countedExecutions++;
+                }
+            }
+        }
+        
+        const totalExecutions = filteredInstances.length;
+        const successfulExecutions = filteredInstances.filter(i => i.status === 'COMPLETED').length;
+        const failedExecutions = filteredInstances.filter(i => i.status === 'ERRORED').length;
```
Fiflon: Maybe increment successfulExecutions and failedExecutions in if (finishedEvents.length > 0)? 
It will decrease looping over filteredInstances.

Fiflon (approved):
one little comment, remember to bump version

runbotics/runbotics-scheduler/src/scheduler-database/process-instance-event/process-instance-event.service.ts
```diff
@@ -90,6 +90,15 @@ export class ProcessInstanceEventService {
         });
     }
 
+    async findAllByProcessInstanceId(processInstanceId: ProcessInstance['id']) {
+        return this.processInstanceEventRepository.find({
+            where: {
+                processInstance: {id: processInstanceId}
```
bhszewczyk: formatting pls ({ id: processInstanceId })

runbotics/runbotics-scheduler/src/utils/process-statistics/process-statistics.service.ts
```diff
@@ -0,0 +1,61 @@
...
+        const averageDuration = countedExecutions > 0 ? totalDuration / countedExecutions : 0;
```
bhszewczyk: for me it sounds like a nice use case for reduce function (instead of multiple filters and push)

also suggestion from me: there is a lot of going on here. can you think of extracting some logic to its own private function(s)? 

consider something like this. looking at this function I should now what is being done:
getProcessStatisticsService
// create date params
// create statistic with these params
// format results

bhszewczyk (changes requested):

bhszewczyk (approved):

PR #1060, opened 2025-07-07
[RPA-2241] Add selfhosted E2E test automation

igorspychalaafo:
## Types of changes

---

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -7,19 +7,17 @@ on:
 
 jobs:
   e2e-tests:
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted]
 
     steps:
-      - name: Checkout E2E Tests
+      - name: Chechout E2E Tests
         uses: actions/checkout@v4
         with:
-          ref: develop
           sparse-checkout: |
             runbotics-test
             .github
 
       - name: Create user.json file
-        working-directory: ./runbotics-test
         run: |
           mkdir -p .auth
           echo '{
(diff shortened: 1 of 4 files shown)


Erquer (approved):

.github/workflows/e2e.yml
```diff
@@ -7,10 +7,10 @@ on:
 
 jobs:
   e2e-tests:
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted]
 
     steps:
-      - name: Checkout E2E Tests
+      - name: Chechout E2E Tests
```
teawithsand: That's a typo

teawithsand (approved):

runbotics-test/tests/example.spec.ts
```diff
@@ -3,22 +3,22 @@ import { readFileSync } from 'node:fs';
...
 
 	await page.getByText('Login', { exact: true }).click()
 
-	await page.waitForURL("/app/processes/collections**");
-	await page.waitForLoadState("load");
+	await page.waitForURL("/app/processes/collections**")
+	await page.waitForLoadState("load")
 
 	expect(await page.title()).toMatch(/RunBotics/i)
 });
```
bhszewczyk: I will be annoying, but can you be consistent with formatting? we have single quotes and double quotes. also - why semicolons were removed? I would go the other way around, so I would add semicolons everywhere :D

I know this is only an example, but maybe it is worth to add different roles here as well (to let also other people who will be creating such tests know that we should consider roles in testing). This is a good example as according to recent views roles USER and TENANT_ADMIN will see "/app/processes/collections**", but role ADMIN will see "/app/tenants**"

igorspychalaafo: This is just a test example — the goal of my task was to create a workflow for automated tests. The Playwright setup, including consistent formatting, semicolons, and role-based test coverage, is being handled by someone else.

PR #1061, opened 2025-07-10
[RPA-2318] Make Firefox the only browser available

teawithsand:
## Types of changes

---


Erquer (approved):
approve, but I don't know if on backlog we weren't talking about preserving chrome options for later development when both browsers will be supported. Comment just to verify, because I'm not sure

bhszewczyk (changes requested):
pls grep chrome in repo and remove/update files accrodingly (e.g. docker-compose, server-config.service.ts)

igorspychalaafo (approved):

bhszewczyk (approved):

PR #1064, opened 2025-07-15
[BLOCKED] API for process summarry notification subscribers

Erquer:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-scheduler/src/scheduler-database/process_summary_notification_subscribers/dto/subscribe.dto.ts
+++ b/runbotics/runbotics-scheduler/src/scheduler-database/process_summary_notification_subscribers/dto/subscribe.dto.ts
@@ -1,0 +1,11 @@
+import { z } from 'zod';
+
+export const subscribeProcessNotificationsDto = z.object({
+    user_id: z.number(),
+    customEmail: z.string().email().optional(),
+    process_id: z.number(),
+});
+
+export type SubscribeDto = z.infer<
+    typeof subscribeProcessNotificationsDto
+>;
--- a/runbotics/runbotics-scheduler/src/scheduler-database/process_summary_notification_subscribers/dto/subscribe.dto.ts
+++ b/runbotics/runbotics-scheduler/src/scheduler-database/process_summary_notification_subscribers/dto/subscribe.dto.ts
@@ -1,11 +1,14 @@
 import { z } from 'zod';
+import { createZodDto } from 'nestjs-zod';
 
 export const subscribeProcessNotificationsDto = z.object({
     user_id: z.number(),
-    customEmail: z.string().email().optional(),
+    customEmail: z.string().trim().email().optional(),
     process_id: z.number(),
 });
 
+
+export class SubscribeSwaggerDto extends createZodDto(subscribeProcessNotificationsDto){}
 export type SubscribeDto = z.infer<
     typeof subscribeProcessNotificationsDto
 >;
--- a/runbotics/runbotics-scheduler/src/scheduler-database/process_summary_notification_subscribers/process-summary-notification-subscribers.controller.ts
+++ b/runbotics/runbotics-scheduler/src/scheduler-database/process_summary_notification_subscribers/process-summary-notification-subscribers.controller.ts
@@ -1,0 +1,55 @@
+import { Body, Controller, Delete, Get, Param, Post, Put } from '@nestjs/common';
+import { ProcessSummaryNotificationSubscribersService } from './process-summary-notification-subscribers.service';
+import { FeatureKeys } from '#/auth/featureKey.decorator';
+import { FeatureKey } from 'runbotics-common';
+import { ZodValidationPipe } from '#/utils/pipes/zod-validation.pipe';
+import {
+    SubscribeDto,
+    subscribeProcessNotificationsDto,
+} from '#/scheduler-database/process_summary_notification_subscribers/dto/subscribe.dto';
+import {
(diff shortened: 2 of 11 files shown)


runbotics/runbotics-scheduler/src/scheduler-database/process_summary_notification_subscribers/process-summary-notification-subscribers.entity.ts
bhszewczyk, line 1: I think table name could be misleading - we already have `notification_process` which is very similar. Maybe better name would be `process_report_subscribers` or something like that? If that was already agreed - disregard my comment.

runbotics/runbotics-scheduler/src/scheduler-database/process_summary_notification_subscribers/dto/subscribe.dto.ts
```diff
@@ -0,0 +1,11 @@
+import { z } from 'zod';
+
+export const subscribeProcessNotificationsDto = z.object({
+    user_id: z.number(),
+    customEmail: z.string().email().optional(),
```
bhszewczyk: you could add trim(), just to be safe

runbotics/runbotics-scheduler/src/scheduler-database/process_summary_notification_subscribers/process-summary-notification-subscribers.controller.ts
```diff
@@ -0,0 +1,55 @@
...
+    async getAllSubscribers() {
+        return this.processNotificationSubscribersService.getAllSubscribersWithProcesses();
+    }
+    
+    @Get('/base-information')
+    async getAllSubscribersBaseInformation() {
+        return this.processNotificationSubscribersService.getAllSubscribersBaseInformation();
+    }
+    
+    @Get('by-process/:processId')
```
bhszewczyk: I think it should be just processes/:processId (to be consistent with other endpoints in scheduler)

runbotics/runbotics-scheduler/src/scheduler-database/process_summary_notification_subscribers/process-summary-notification-subscribers.service.ts
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    private async sendAggregatedStatisticsEmail(
+        email: string,
+        summaries: { name: string, stats: ProcessStatisticsResult }[],
+    ) {
+        const htmlContent = this.generateAggregatedEmailContent(summaries);
+        console.log(`Sending aggregated email to ${email}`);
+        console.log(`Sending aggregated email with content: ${htmlContent}`);
```
bhszewczyk, line 62: use logger here

runbotics/runbotics-scheduler/src/scheduler-database/process_summary_notification_subscribers/process-summary-notification-subscribers.service.ts
```diff
@@ -0,0 +1,130 @@
...
+                <li>Total Executions: ${stats.totalExecutions}</li>
+                <li>Successful Executions: ${stats.successfulExecutions}</li>
+                <li>Failed Executions: ${stats.failedExecutions}</li>
+                <li>Average Duration (seconds): ${stats.averageDuration}</li>
+                <li>Total Duration (seconds): ${stats.totalDuration}</li>
+            </ul>
+        `).join('');
+
+        return `<h1>Process Summary</h1>${content}`;
+    }
```
bhszewczyk, line 84: we do not want to have the same in Polish?

runbotics/runbotics-scheduler/src/scheduler-database/process_summary_notification_subscribers/process-summary-notification-subscribers.controller.ts
bhszewczyk, line 1: use GetWithTenant decorator and UserDecorator (check e.g. licence controller or user controller). We need to make sure the person who sends the request is under the proper tenant and only then we can proceed. I know the processId is unique, but it is just an int so easy to guess. From the code I understand person from different tenant having the right REST url will get information which suppose to be not found for him via API without any check if the user who requests it is under the tenant in the path or not.

bhszewczyk (changes requested):

runbotics/runbotics-scheduler/src/scheduler-database/process_summary_notification_subscribers/process-summary-notification-subscribers.service.ts
Erquer: it's not my code, this function is part of different task. (PR is not yet created)

runbotics/runbotics-scheduler/src/scheduler-database/process_summary_notification_subscribers/process-summary-notification-subscribers.service.ts
Erquer: as above

runbotics/runbotics-scheduler/src/scheduler-database/process_summary_notification_subscribers/process-summary-notification-subscribers.entity.ts
Erquer: not task related comment - the name was discussed with PO as far as I know

igorspychalaafo (approved):

bhszewczyk (approved):

PR #1085, opened 2025-08-06
[RPA-2360] add Active and subEnd columns to tenant

Erquer:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-scheduler/src/scheduler-database/tenant/tenant.entity.ts
+++ b/runbotics/runbotics-scheduler/src/scheduler-database/tenant/tenant.entity.ts
@@ -18,6 +18,7 @@ import { Credential } from '../credential/credential.entity';
 import { User } from '../user/user.entity';
 import { EmailTriggerWhitelistItem } from '../email-trigger-whitelist-item/email-trigger-whitelist-item.entity';
 import { ApiProperty } from '@nestjs/swagger';
+import { dateTransformer } from '#/database/database.utils';
 
 @Entity()
 export class Tenant {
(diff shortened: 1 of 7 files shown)


runbotics/runbotics-scheduler/src/scheduler-database/tenant/tenant.entity.ts
```diff
@@ -81,4 +81,12 @@ export class Tenant {
...
     )
     emailTriggerWhitelist: EmailTriggerWhitelistItem[];
+
+    @ApiProperty({ type: 'boolean', default: true })
+    @Column({ type: 'boolean', nullable: false, default: true })
+    active: boolean;
+
+    @ApiProperty({ type: 'string', format: 'date-time', nullable: true })
+    @Column({ type: 'timestamp', nullable: true })
+    subscriptionEnd: Date | null;
```
bhszewczyk: I think this should be adjusted to the date format we have in our database, e.g. `process-instance.entity.ts`. If not the case and it is not needed this disregard, but worth to mention we have transformer for dates.
```
    @Column({
        type: 'timestamp without time zone',
        transformer: dateTransformer,
        nullable: true,
    })
    updated?: string;
```

bhszewczyk (changes requested):

runbotics/runbotics-scheduler/src/scheduler-database/tenant/tenant.entity.ts
Erquer: good point, thank you

bhszewczyk (approved):
bump gradle version

sniperdev (approved):

PR #1089, opened 2025-08-08
Global Guard for validatig active tenant

Erquer:
## Description

add to cheduler global guard for checking if tenant is active

## Motivation and Context

in case, user is logged and tenant will become inactive, we should block requests for resources related to this tenant

## Types of changes

---


Fiflon (approved):
Bump version :D

sniperdev (approved):

PR #1096, opened 2025-08-18
bugfix on blacklist

Erquer:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-scheduler/src/queue/processor/scheduler.processor.ts
+++ b/runbotics/runbotics-scheduler/src/queue/processor/scheduler.processor.ts
@@ -56,12 +56,12 @@ export class SchedulerProcessor {
         const isAllowedToRun = !(await this.blacklistActionAuthService.checkProcessActionsBlacklist(process.id ?? job.data.process.id));
 
         if(!isAllowedToRun) {
-            this.logger.error(`[Q Process] Process "${process.name}" is blacklisted and cannot be run.`);
+            this.logger.error(`[Q Process] Process "${process.name}" contains blacklisted actions and cannot be run.`);
             await this.processInstanceSchedulerService.saveFailedProcessInstance(
                 job,
-                'Process is blacklisted',
+                'Process contains blacklisted actions',
             );
-            throw new BadRequestException(`Process ${process.id} is blacklisted`);
+            throw new BadRequestException(`Process ${process.id} contains blacklisted actions and cannot be run.`);
         }
         if (!isScheduledProcess(job.data)) {
             return this.runProcess(process, job);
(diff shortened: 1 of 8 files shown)


runbotics/runbotics-scheduler/src/queue/processor/scheduler.processor.ts
```diff
@@ -50,23 +50,28 @@ export class SchedulerProcessor {
...
-        const isAllowedToRun = await this.blacklistActionAuthService.checkProcessActionsBlacklist(job.data.process.id);
-
         const process = await this.processService.findById(job.data.process.id);
-        if (!isScheduledProcess(job.data) && isAllowedToRun) {
+
+        //reverse value of check, because it returns if the process is blacklisted, so is allowed when is not blacklisted
+        const isAllowedToRun = !(await this.blacklistActionAuthService.checkProcessActionsBlacklist(process.id ?? job.data.process.id));
+
+        if(!isAllowedToRun) {
+            this.logger.error(`[Q Process] Process "${process.name}" is blacklisted and cannot be run.`);
```
bhszewczyk: can whole process be blacklisted? If yes - that's ok, but if the process cannot run because it includes blacklisted actions I think the message should say so.

bhszewczyk (changes requested):
Please change branch name to bugfix/RPA-2353 (or other task number if more appropriate) and change the PR name as well - we wouldn't want to have it like this when we do a next release

runbotics/runbotics-scheduler/src/queue/processor/scheduler.processor.ts
Fiflon: If we do not want to disclose which actions are blacklisted (for some reason), we can simply write that the process contains prohibited actions.

runbotics/runbotics-scheduler/src/queue/processor/scheduler.processor.ts
```diff
@@ -50,23 +50,28 @@ export class SchedulerProcessor {
...
+            await this.processInstanceSchedulerService.saveFailedProcessInstance(
+                job,
+                'Process is blacklisted',
```
Fiflon: To me it should inform that process contains blacklisted actions

Fiflon (approved):
small comments

runbotics/runbotics-scheduler/src/queue/processor/scheduler.processor.ts
Erquer: function return just boolean, imo I can change message to process is failed because contains blacklisted actions.

runbotics/runbotics-scheduler/src/queue/processor/scheduler.processor.ts
Erquer: function return just boolean, imo I can change message to process is failed because contains blacklisted actions.

Fiflon:
> Please change branch name to bugfix/RPA-2353 (or other task number if more appropriate) and change the PR name as well - we wouldn't want to have it like this when we do a next release

+1

PR #1108, opened 2025-08-27
[RPA-2419] Add new role ROLE_RPA_USER and change usage of ROLE_USER

Fiflon:
## Description

I added migrations that convert ROLE_USER to ROLE_RPA_USER.
I changed the enums to support the new role as the old ROLE_USER.
I added additional security to the assistant.
I added a new role, ROLE_USER.
I modified the enums again.
I added the new ROLE_USER as the default role after SSO.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

---


teawithsand (approved):

runbotics/runbotics-orchestrator-ui/src/src-app/components/guards/GuestGuard.tsx
```diff
@@ -25,16 +25,28 @@ export const withGuestGuard = (Component: FC | VFC) => (props: any) => {
...
         }
 
-        if (![Role.ROLE_GUEST, Role.ROLE_ADMIN].some(role => user.roles.includes(role))) {
-            router.replace('/app/processes/collections', null, { locale: user.langKey });
+        if (isJustUser) {
+            router.replace('/app/assistant', null, { locale: user.langKey });
         }
 
+        if (
+            !isJustUser &&
```
janeksmielowski: maybe we can go for `else if` and then skip the first condition, because it's asserted by the first `if` statement?

runbotics/runbotics-orchestrator-ui/src/src-app/layouts/MainLayout/NavBar/usePublicSections.ts
```diff
@@ -105,7 +105,7 @@ export const usePublicSections = (): Section[] => {
                     title: translate('Nav.Items.AIAssistant'),
                     icon: SmartToyOutlinedIcon,
                     href: '/app/assistant',
-                    authorities: [Role.ROLE_USER], // todo: change Role.ROLE_USER to the 'new' ROLE_USER when it appears
+                    authorities: [Role.ROLE_USER],
```
janeksmielowski, line 108: I understand that this is **expected** not to be changed, because /app/assistant is only authorized for ROLE_USER. Please confirm

runbotics/runbotics-orchestrator-ui/src/src-app/layouts/MainLayout/TopBar/index.tsx
```diff
@@ -76,10 +87,13 @@ const TopBar: FC<TopBarProps> = ({ className, ...rest }) => {
     const isTenantAdmin = useRole([Role.ROLE_TENANT_ADMIN]);
     const isAdmin = useRole([Role.ROLE_ADMIN]);
     const isGuest = user?.roles.includes(Role.ROLE_GUEST);
+    const isJustUser = user?.roles.every((role) => role === Role.ROLE_USER);
```
janeksmielowski: couldn't we create a helper hook that will return all of these? Or extend useAuth to actually return: `isTenantAdmin`, `isAdmin`, `isGuest`, `isJustUser` ?

janeksmielowski: btw. I don't like `isJustUser`. Go for `isUser` and `isRpaUser` if needed somewhere else

runbotics/runbotics-orchestrator-ui/src/src-app/views/AIAssistantView/AIAssistantView.tsx
```diff
@@ -1,15 +1,28 @@
...
 
+
 const AIAssistantView: VFC = () => {
     const { translate } = useTranslations();
+    const { user } = useAuth();
+
 
     const pageTitle = translate('AIAssistant.Title');
 
+    const isTenantDefault = user.tenant.id === 'b7f9092f-5973-c781-08db-4d6e48f78e98'; // temporary additional check
```
janeksmielowski, line 20: This is fine - we will remove it in the future, if we decide to support other tenants

runbotics/runbotics-scheduler/src/migrations/1755157999999-AlterUserRoleWithRpaUser.ts
```diff
@@ -0,0 +1,23 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+ 
+export class AlterUserRoleWithRpaUser1755157999999
+    implements MigrationInterface
+{
+    name = 'AlterUserRoleWithRpaUser1755157999999';
+ 
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`
+            UPDATE "jhi_authority"
```
janeksmielowski, line 10: please make sure this is the only table to be updated

janeksmielowski (approved):

runbotics/runbotics-orchestrator-ui/src/src-app/views/AIAssistantView/AIAssistantView.tsx
janeksmielowski: one question: is this tenant ID common in all evironments (local, dev, qas, prod, com)?

runbotics/runbotics-orchestrator-ui/src/src-app/components/guards/GuestGuard.tsx
Fiflon: Great idea

runbotics/runbotics-orchestrator-ui/src/src-app/layouts/MainLayout/NavBar/usePublicSections.ts
Fiflon: Yes, it used to be 'old' ROLE_USER. Now it's 'new' ROLE_USER

runbotics/runbotics-scheduler/src/migrations/1723195979632-CredentialsFeatureKey.ts
bhszewczyk, line 1: for the future: NEVER update migration which is already in executed
here you use ROLE_RPA_USER which is not in the database yet.
the flow for swap roles should be like this: rename existing role to the new role -> new role get all the feature keys which were previously assigned to the 'old'role -> if change in feature keys needed in a new role - adjust it -> now we should have state in db where there is no old role => then you can re-create new role and use the same role as old one and add feature keys which are applicable for the new role

runbotics/runbotics-scheduler/src/migrations/1755157999999-AlterUserRoleWithRpaUser.ts
bhszewczyk, line 1: so here you add the ROLE_RPA_USER so how it is possible you use it before adding such role?

runbotics/runbotics-scheduler/src/migrations/1725274222207-ActionFeatureKey.ts
bhszewczyk, line 1: to revert

runbotics/runbotics-scheduler/src/migrations/1735571296131-TagFeatureKey.ts
bhszewczyk, line 1: to revert

runbotics/runbotics-scheduler/src/migrations/1755157999999-AlterUserRoleWithRpaUser.ts
bhszewczyk: to be sure: 
1. the table where user is connected to role is also changed?
2. the table in which we have feature keys assigned to the role will also be updated?

this is why I suggested to make it in 2 steps: to be sure it does not affect current users

runbotics/runbotics-common/src/utils/getTenantAllowedRoles.ts
```diff
@@ -0,0 +1,3 @@
+import { Role } from "../model";
+
+export const getTenantAllowedRoles = () => [Role.ROLE_RPA_USER, Role.ROLE_USER, Role.ROLE_EXTERNAL_USER, Role.ROLE_TENANT_ADMIN];
```
bhszewczyk, line 3: I would suggest slightly new name getRolesAllowedInTenant, because from the current name I was not sure if it refers to tenant or to tenant role

runbotics/runbotics-orchestrator-ui/src/src-app/components/guards/GuestGuard.tsx
```diff
@@ -25,16 +25,28 @@ export const withGuestGuard = (Component: FC | VFC) => (props: any) => {
...
     if (isBrowser && isInitialized && isAuthenticated) {
+        const isJustUser = user.roles.every((role) => role === Role.ROLE_USER);
         if (user.roles.includes(Role.ROLE_ADMIN)) {
             router.replace('/app/tenants', null, { locale: user.langKey });
         }
 
-        if (![Role.ROLE_GUEST, Role.ROLE_ADMIN].some(role => user.roles.includes(role))) {
-            router.replace('/app/processes/collections', null, { locale: user.langKey });
+        if (isJustUser) {
+            router.replace('/app/assistant', null, { locale: user.langKey });
```
bhszewczyk, line 36: if you add return here you don't need to check below if !isJustUser

runbotics/runbotics-orchestrator-ui/src/src-app/views/AIAssistantView/AIAssistantView.tsx
bhszewczyk: yes, I can confirm it is the same for all envs

bhszewczyk (changes requested):

PR #1119, opened 2025-09-03
[RPA-2427] Fix SSO bug

Fiflon:
## Description

I changed email given from sso to be lowercase and trimmed it for safety.

## Types of changes

---

### DIFF
--- a/runbotics/runbotics-scheduler/src/scheduler-database/msal/msal.service.ts
+++ b/runbotics/runbotics-scheduler/src/scheduler-database/msal/msal.service.ts
@@ -163,7 +163,7 @@ export class MsalService {
         }
 
         const authToken = await this.authService.handleMsalSsoAuth({
-            email: profileData.email,
+            email: profileData.email.toLowerCase().trim(),
             langKey: profileData.langKey,
             msTenantId: loginResponse.tenantId,
             msObjectId: loginResponse.objectId,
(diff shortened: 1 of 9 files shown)


runbotics/runbotics-scheduler/src/scheduler-database/msal/msal.service.ts
```diff
@@ -163,7 +163,7 @@ export class MsalService {
         }
 
         const authToken = await this.authService.handleMsalSsoAuth({
-            email: profileData.email,
+            email: profileData.email.toLowerCase().trim(),
```
Erquer, line 166: you forgot about creating ms user, checking there if user exists and only fill microsoft fields for existing user. I suppose it was in user.service.ts(?)

Erquer (changes requested):

runbotics/runbotics-scheduler/src/scheduler-database/msal/msal.service.ts
Fiflon: No, this mechanism is already implemented. The problem was that it checked whether exactly the same email address existed in the database (without converting to lowercase, so if the email address was already in use and the new address contained capital letters, it treated it as a new email address).

Erquer (approved):
remember bump version.

bhszewczyk (approved):

PR #1145, opened 2025-09-22
[RPA-2446] JWT Token generation and authentication via Scheduler not Orchestrator API

Erquer:
## Types of changes

---


runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Auth/Auth.thunks.ts
```diff
@@ -17,53 +17,59 @@ export const setAccessToken = (accessToken: string | null): void => {
...
+export const createGuestAccount = createAsyncThunk<
+    UserDto,
+    { langKey: string }
+>('auth/createGuestAccount', async ({ langKey }, { rejectWithValue }) => {
+    try {
+        const response = await Axios.post<{ idToken: string; user: UserDto }>(
+            '/api/scheduler/auth/authenticate/guest',
+            { langKey }
+        );
+        const token = response.data.idToken;
```
bhszewczyk: you can use here exactly the same logic as for login (I mean `const { idToken, user } = response.data;`) for consistency and simplification and even extract returned object to another function to make it DRY

runbotics/runbotics-orchestrator-ui/src/src-app/store/slices/Auth/Auth.thunks.ts
```diff
@@ -17,53 +17,59 @@ export const setAccessToken = (accessToken: string | null): void => {
...
-                {
-                    username: payload.email,
-                    password: payload.password,
-                    rememberMe: true,
-                }
-            );
-            const { id_token: idToken } = response.data;
-
-            setAccessToken(idToken);
-            const responseUser = await Axios.get<UserDto>('/api/account');
```
bhszewczyk, line 35: just to make sure - now enpoint /authenticate returns also user - I am not sure if that is a good solution. let's discuss it

runbotics/runbotics-scheduler/src/auth/auth.service.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+    describe('authenticate', () => {
+        const baseUser = {
+            email: 'x@test.com',
+            passwordHash: 'hashed',
+            authorities: [{ name: 'ROLE_USER' }],
+            tenant: { active: true },
+            activated: true,
+        };
+
+        it('✅ poprawne dane zwracają token i użytkownika', async () => {
```
Fiflon: Maybe translate te English and remove emoji?

runbotics/runbotics-scheduler/src/auth/dto/auth.dto.ts
```diff
@@ -0,0 +1,12 @@
+import { z } from 'zod';
+
+export const authSchema = z.object({
+    username: z.string().trim().min(1).max(100),
+    password: z.string()
+        .trim()
+        .min(14, 'Password must be at least 14 characters long')
```
bhszewczyk, line 8: let's make min characters a const variable

runbotics/runbotics-scheduler/src/auth/auth-guest.service.spec.ts
bhszewczyk, line 1: you had to expect that xD: please change describe messages to English :D

runbotics/runbotics-scheduler/src/auth/auth-guest.service.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+        it('powinno zwrócić true, gdy nie ma pasujących gości', async () => {
+            dataSourceMock.manager.find.mockResolvedValue([
+                { ipHash: 'hash1' },
+                { ipHash: 'hash2' },
+            ]);
+            (bcrypt.compareSync as any).mockReturnValue(false);
+
+            const result = await service.verifyGuestLimit('127.0.0.1');
+            expect(result).toBe(true);
+            expect(dataSourceMock.manager.find).toHaveBeenCalledWith(Guest);
```
bhszewczyk, line 48: why here we do have this line but in the next test we do not?

runbotics/runbotics-scheduler/src/auth/auth.service.spec.ts
bhszewczyk, line 1: remove emojis and change it to English pls :D

runbotics/runbotics-scheduler/src/auth/auth.service.ts
```diff
@@ -328,5 +336,73 @@ export class AuthService {
...
+        return {
+            idToken: this.jwtService.sign(payload, {
+                expiresIn: rememberMe ? '30d' : '1d',
+                algorithm: 'HS512',
+                noTimestamp: true,
+            }),
+            user: {
+                ...user,
+                authorities: undefined,
+                passwordHash: undefined,
```
bhszewczyk: for returning both token and user: tbh I preferred previous solution where /authenticate returned only token and /account returned user. seems like violation of the single responsibility rule

for authorities and passwordHash undefined setup: I would use explicit pick rather than setting some attributes to undefined. also - I am wondering if better option would be to use serializer here. E.g.

```
class UserSafeDto {
    constructor(user: User) {
        this.id = user.id;
        this.email = user.email;
        // ...other allowed fields
    }
    id: string;
    email: string;
    // add allowed fields only
}
```

and use it here as:

```
return {
    idToken: // the same as it is now
    user: new UserSafeDto(user),
};
```

runbotics/runbotics-scheduler/src/auth/auth.service.ts
```diff
@@ -328,5 +336,73 @@ export class AuthService {
...
+                expiresIn: '1h',
+                algorithm: 'HS512',
+                noTimestamp: true,
+            }),
+            user: {
+                ...user,
+                authorities: undefined,
+                passwordHash: undefined,
+            },
+        };
```
bhszewczyk: same comment here as for authenticate

runbotics/runbotics-scheduler/src/scheduler-database/user/user.controller.ts
```diff
@@ -37,6 +37,15 @@ export class UserController {
...
     }
 
+    @Get('users/account')
+    @FeatureKeys(FeatureKey.TENANT_READ_USER, FeatureKey.PROCESS_ADD_GUEST)
+    getUserByIdInTenant(
+        @UserDecorator() { email }: User,
+    ) {
+        return this.userService.findByEmailForAuth(email);
+        
+    }
```
bhszewczyk: so as user A I can access this endpoint, provide e-mail of user B and get the hashedPassword? 😨 hope i miss something here :D

runbotics/runbotics-scheduler/src/scheduler-database/user/user.service.ts
```diff
@@ -135,13 +142,42 @@ export class UserService {
...
+        if(!result) {
+            throw new NotFoundException(`User with email ${email} not found`);
+        }
+        const password = await this.userRepository.findOne({
+           select: { email:true, id: true, passwordHash: true},
+            where: { id: result.id },
+        });
+        return {
+            ...result, 
+            passwordHash: password.passwordHash,
```
bhszewczyk, line 165: I think that should never be exposed in api and now we have situation where I can get the passwordHash when i get endpoint /users/acciunt?

bhszewczyk (changes requested):

runbotics/runbotics-scheduler/src/auth/auth-guest.service.spec.ts
Erquer: because in next test we do not reach find function in code.

runbotics/runbotics-scheduler/src/auth/auth.controller.ts
Fiflon, line 1: Remember to add swagger to the new controllers ;)

runbotics-orchestrator/src/main/java/com/runbotics/config/LocaleConfiguration.java
```diff
@@ -28,8 +28,6 @@ public void addInterceptors(InterceptorRegistry registry) {
         localeChangeInterceptor.setParamName("language");
         registry.addInterceptor(localeChangeInterceptor);
         registry.addInterceptor(accountInterceptor).excludePathPatterns(
-            "/api/authenticate/guest",
-            "/api/authenticate",
             "/api/register"
```
janeksmielowski, line 31: what about register?

runbotics/runbotics-orchestrator-ui/src/src-app/utils/axios.ts
```diff
@@ -28,7 +28,7 @@ axiosInstance.interceptors.response.use(
 const axiosApi = axios.create();
 
 axiosApi.interceptors.response.use((res) => res, (err) => {
-    if (err.request.status === 401 && !err.request.responseURL.includes('/api/authenticate')) {
+    if (err.request.status === 401 && !err.request.responseURL.includes('/api/scheduler/auth/authenticate')) {
```
janeksmielowski: shouldn't it stay as we discussed? /api/authenticate should be now proxied to scheduler

runbotics/runbotics-proxy/proxy-server.js
```diff
@@ -24,24 +24,28 @@ const hosts = {
...
+    {path: '/register', target: hosts.ui, pathFilter: ['/register/**', '!/pl/register/**']},
+    {path: '/ui', target: hosts.ui, pathFilter: ['/ui/**', '!/pl/ui/**']},
+    {path: '/pl/ui', target: hosts.ui, pathFilter: '/pl/ui/**'},
+
+    {path: '/api/account', target: hosts.scheduler, pathFilter: '/api/account'},
+    {path: '/api/authenticate', target: hosts.scheduler, pathFilter: '/api/authenticate'},
+    {path: '/api/authenticate/guest', target: hosts.scheduler, pathFilter: '/api/authenticate/guest'},
+
+    {path: '/api/scheduler', target: hosts.scheduler, pathFilter: '/api/scheduler/**'},
+    {path: '/api', target: hosts.api, pathFilter: ['/api/**', '!/api/account', '!/api/authenticate', '!/api/authenticate/guest','!/api/scheduler/**']},
```
bhszewczyk, line 44: can't we now just `{path: '/api', target: hosts.scheduler, pathFilter: '/api/**'}`? 🤔

bhszewczyk (approved):

runbotics/runbotics-proxy/proxy-server.js
bhszewczyk: as discussed - not yet 😭

runbotics/runbotics-scheduler/src/auth/auth.controller.ts
```diff
@@ -0,0 +1,81 @@
...
+            }
+        }})
+    @ApiResponse({ status: 403, description: 'Guest limit exceeded' })
+    async authenticateGuest(
+        @Body() body: { langKey: string },
+        @Req() req: Request,
+        @Res() res: Response,
+    ) {
+        this.logger.log('Authenticating guest user');
+        const firstForwardedFor = Array.isArray(req.headers['x-forwarded-for']) ?
```
teawithsand, line 67: This may not be safe and may let user provide arbitrary value of ip here.

For example: imagine I've sent a request with x-forwarded-for: 1.2.3.4 from my ip of 1.1.1.1 x-forwarded-for is header like any other one.

Then IT's nginx would append my real IP(todo: verify how IT's nginx behaves; since this depends on nginx config file) and RB nginx would append IT's nginx IP(it does so at the moment of writing this), giving total of N+2 IPs in this header, where N is number of IPs provided in initial request.

This is not a code issue per se, but requires additional configuration on reverse proxy side to handle that. Clearing of initial value of x-forwarded-for is required. Perhaps X-Real-IP can be reconfigured not to point to IT's nginx, as it does right now afaik.

teawithsand (approved):

runbotics-test/pages/loginPage.ts
```diff
@@ -22,6 +24,14 @@ export class LoginPage {
...
+    async loginGuest() {
+        await this.loginGuestButton.click();
+    }
+    
+    async waitForLoginGuestSuccess() {
+        await this.page.waitForURL("**/app/processes/**");
+    }
 
     async waitForLoginSuccess() {
         await this.page.waitForURL("**/app/processes/collections**");
```
Fiflon, line 37: Maybe its a good idea to refactor this code to sth like waitForLoginSuccess(role: some enum with paths related to the role after login)? I'm afraid that we will end up with 4 same functions with little to no differences.

runbotics/runbotics-orchestrator-ui/src/src-app/hooks/useAuth.ts
Fiflon, line 1: imo great memo addition!

janeksmielowski (approved):

runbotics/runbotics-scheduler/src/auth/auth.controller.ts
Fiflon: right now, somebody can make limitless amount of guest accounts :o

Fiflon (approved):
LGTM

PR #1146, opened 2025-09-23
[RPA-2468] add validation of tenant and validation of password while …

Erquer:
…login

## Types of changes

---

### DIFF
--- a/runbotics/runbotics-scheduler/src/auth/auth.controller.ts
+++ b/runbotics/runbotics-scheduler/src/auth/auth.controller.ts
@@ -23,7 +23,7 @@ export class AuthController {
         @Body(new ZodValidationPipe(authSchema)) body: AuthDto,
         @Res() res: Response,
     ) {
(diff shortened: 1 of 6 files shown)


runbotics/runbotics-scheduler/src/auth/auth.controller.ts
```diff
@@ -14,10 +16,10 @@ export class AuthController {
     @Public()
     @Post('authenticate')
     async authenticate(
-        @Body() body: { username: string; password: string, rememberMe?: string },
+        @Body(new ZodValidationPipe(authSchema)) body: AuthDto,
         @Res() res: Response,
     ) {
-        const newToken = await this.authService.authenticate(body.username, body.password, body.rememberMe === 'true');
+        const newToken = await this.authService.authenticate(body.username, body.password, body.rememberMe);
```
sniperdev: Wouldn't it be better to pass one object instead of three variables?

sniperdev (approved):

PR #1151, opened 2025-10-02
[RPA-2408] Fixes bug, when user could access bot details (/logs, /configure) by manipulating url

bhszewczyk:
## Description

Previously user could access bot details by changing id in url for bot logs and bot configure
Added check if user who sends the request has access to the bot at all on the details level
Now on frontend there is redirect to 404 if the user does not have access to the bot and try to enter bot logs or bot configure tabs via url directly
On backend endpoint for process-instances/GetPage is also secured

## Types of changes

---


runbotics/runbotics-scheduler/src/scheduler-database/process-instance/process-instance.controller.ts
```diff
@@ -34,13 +36,17 @@ export class ProcessInstanceController {
...
-    getPage(
+    async getPage(
         @Specifiable(ProcessInstanceCriteria) specs: Specs<ProcessInstance>,
         @Pageable() paging: Paging,
         @UserDecorator() user: User,
         @Query('processId.equals') processId: string,
         @Query('botId.equals') botId: string,
     ) {
+        if (botId) {
+            await this.botCrudService.findOne(user, parseInt(botId));
```
Fiflon, line 47: I think it would be best to add some kind of validation error like: 
```
        const parsedBotId = parseInt(botId, 10);
        
        if (isNaN(parsedBotId) || parsedBotId <= 0) {
            throw new BadRequestException('botId must be a valid positive number');
        }
```

runbotics/runbotics-orchestrator-ui/src/src-app/views/bot/BotDetailsView/index.tsx
```diff
@@ -30,24 +28,56 @@ const StyledPage = styled(Page)(({ theme }) => ({
...
+            <StyledPage className={classes.root} title={translate('Bot.Details.Meta.Title')}>
+                <Container maxWidth={false}>
+                    <Loader/>
+                </Container>
+            </StyledPage>
+        );
+    }
+
+    if (isError) {
+        return null;
```
Fiflon, line 78: Maybe display some kind of information when errored? I think it would look good.

runbotics/runbotics-scheduler/src/scheduler-database/process-instance/process-instance.controller.ts
Fiflon: What do you think about creating a function that returns Boolean or void? We do nothing bot that findOne returns.
Maybe we should create guard??? Like 
```
@Injectable()
export class BotAccessGuard implements CanActivate {
    constructor(private botCrudService: BotCrudService) {}

    async canActivate(context: ExecutionContext): Promise<boolean> {
        const request = context.switchToHttp().getRequest();
        const user = request.user;
        const botId = request.query['botId.equals'];

        try {
            const parsedBotId = parseInt(botId, 10);
            if (isNaN(parsedBotId)) {
                throw new BadRequestException('Invalid botId format');
            }

            await this.botCrudService.findOne(user, parsedBotId);
            return true;
        } catch (error) {
            if (error instanceof NotFoundException) {
                throw new BadRequestException('Bot not found or not accessible');
            }
            throw error;
        }
    }
}
```
Maybe im just overthinking this... You do you ;)

Fiflon (approved):
All good but i have couple suggestions.

runbotics/runbotics-scheduler/src/scheduler-database/process-instance/process-instance.controller.ts
bhszewczyk: Guard is a good idea, but to use it only in one place not so much. In the future when we will have new process collections I think it will be nice to refactor accesses to bots as well using similar strategy patterns. Validation error is not needed imo here, it is not provided anywhere by the user so in case it is being called in the code and there is no such bot we just give 403 or 404 and that is enough.

runbotics/runbotics-orchestrator-ui/src/src-app/views/bot/BotDetailsView/index.tsx
bhszewczyk: Not needed, if error there is 404 redirect

Erquer (approved):

PR #1153, opened 2025-10-02
[RPA-2412] Bump version to sync changes from gitea

teawithsand:
On the gitea side, there were some version bumps, which were required in order to trigger auto deployment there.

Since these versions were pushed to docker registry, a new version bump is required here on GH.

Note that all changes from gitea can't be just ported to GH, since these include changes to GH actions/workflows, which in turn would break them on github, as they were adapted for gitea.

### DIFF
--- a/runbotics-orchestrator/build.gradle
+++ b/runbotics-orchestrator/build.gradle
@@ -27,7 +27,7 @@ plugins {
 group = "com.runbotics"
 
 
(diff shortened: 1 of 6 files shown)


Erquer (approved):

bhszewczyk (approved):

Fiflon (approved):

Fiflon:
lgtm <3

PR #1164, opened 2025-10-13
[RPA-2452] Translations in emails

sniperdev:
## Types of changes

---


runbotics/runbotics-scheduler/src/mail/i18n.service.ts
```diff
@@ -0,0 +1,67 @@
+import { Logger } from '#/utils/logger';
+import { Injectable } from '@nestjs/common';
+import mailEn from './i18n/en/mail.json';
+import mailPl from './i18n/pl/mail.json';
+
+interface TranslationObject {
+    [key: string]: string | TranslationObject;
+}
+
+type SupportedLanguage = 'en' | 'pl';
```
bhszewczyk: If not there yet I think this should be defined in common

runbotics/runbotics-scheduler/src/mail/i18n.service.ts
```diff
@@ -0,0 +1,67 @@
...
+                pl: mailPl as TranslationObject,
+            };
+            this.logger.log(`I18n loaded, langs: ${Object.keys(this.translations).join(', ')}`);
+        } catch (err: unknown) {
+            const errorMessage = err instanceof Error ? err.message : String(err);
+            this.logger.error(`I18n initialization failed: ${errorMessage}`);
+        }
+    }
+
+    translate(key: string, lang = 'en', vars: Record<string, string> = {}): string {
```
bhszewczyk: here 'en' should be a const DEFAULT_LANGUAGE and also it could be moved to common

runbotics/runbotics-scheduler/src/mail/i18n.service.ts
```diff
@@ -0,0 +1,67 @@
...
+            } else {
+                return undefined;
+            }
+        }
+        
+        return typeof current === 'string' ? current : undefined;
+    }
+
+    private interpolate(text: string, vars: Record<string, string>): string {
+        if(typeof text !== 'string') return text;
```
bhszewczyk: i don't get it, you assigned type to text to be string and defined this function to return string and here you check text (which you typed to string) and return not string?

runbotics/runbotics-scheduler/src/mail/i18n/en/mail.json
```diff
@@ -0,0 +1,58 @@
...
+            "copyright": "{{year}} © {{env}} v{{version}}"
+        },
+        "credentialChange": {
+            "subject": "Credential change inside owned collection",
+            "messageWithAttribute": "User with email <b>{{editorEmail}}</b> {{operationType}} with name <i>{{attributeName}}</i> for credential <b>{{credentialName}}</b>{{oldNameInfo}} in collection <b>{{collectionName}}</b>",
+            "messageWithoutAttribute": "User with email <b>{{editorEmail}}</b> {{operationType}} credential <b>{{credentialName}}</b>{{oldNameInfo}} in collection <b>{{collectionName}}</b>",
+            "oldNameInfo": " (name before change <b>{{oldName}}</b>)"
+        },
+        "userActivation": {
+            "subject": "RunBotics - User Activation"
```
bhszewczyk, line 56: no email body here? :D

runbotics/runbotics-scheduler/src/mail/i18n.service.ts
```diff
@@ -0,0 +1,67 @@
...
+
+        return this.interpolate(value, vars);
+    }
+
+    private getNestedValue(obj: TranslationObject, path: string): string | undefined {
+        const parts = path.split('.');
+        let current: string | TranslationObject | undefined = obj;
+        
+        for (const part of parts) {
+            if (typeof current === 'object' && current !== null && part in current) {
```
bhszewczyk: Why we do not return null instead of undefined?

I suggest slight change: `(current && typeof current === 'object' && part in current)` or even early guard before if)

if (!current) return null

bhszewczyk: for me this function name is to general. what value? I suggest to rename it to something more clear like e.g. getTranslation

runbotics/runbotics-scheduler/src/mail/i18n.service.ts
bhszewczyk: I would rename functions as follows:
translate -> buildTranslation,
getNestedValue -> getTranslation
interpolate-> translate

If I didn't get something or you have other suggestions let me know, I just don't know what exactly getNestedValue does without reading all the code

runbotics/runbotics-scheduler/src/mail/mail.service.ts
```diff
@@ -99,7 +95,25 @@ export class MailService {
...
-        await this.handleNotificationEmail(sendMailInput, [botAssignedUserEmail, ...targetEmails]);
+        const allEmails = [botAssignedUserEmail, ...targetEmails];
+
+        for (const email of allEmails) {
+            const userLang = await this.getUserLanguage(email);
+            const subject = this.i18n.translate('mail.botDisconnection.subject', userLang);
+            const greeting = this.i18n.translate('mail.botDisconnection.greeting', userLang);
+            const message = this.i18n.translate('mail.botDisconnection.botDisconnectedMessage', userLang, { installationId });
+            const visitLink = this.i18n.translate('mail.botDisconnection.visitLink', userLang);
+            const signature = this.i18n.translate('mail.botDisconnection.signature', userLang);
```
bhszewczyk: I think this should be a separate function named getBotDisconnectionEmail

runbotics/runbotics-scheduler/src/mail/mail.service.ts
```diff
@@ -123,10 +131,30 @@ export class MailService {
...
+
+        for (const email of allEmails) {
+            const userLang = await this.getUserLanguage(email);
+            const subject = this.i18n.translate('mail.processFailure.subject', userLang);
+            const greeting = this.i18n.translate('mail.processFailure.greeting', userLang);
+            const message = this.i18n.translate('mail.processFailure.processFailedMessage',  userLang, 
+                { processName: process.name, processId: process.id.toString(), status: processInstance.status } 
+            );
+            const visitLink = this.i18n.translate('mail.processFailure.visitLink', userLang);
+            const signature = this.i18n.translate('mail.processFailure.signature', userLang);
```
bhszewczyk: extract to getProcessFailureEmail

runbotics/runbotics-scheduler/src/mail/templates/i18n/subscription-expiration-notification.template.ts
```diff
@@ -0,0 +1,54 @@
...
+    </style>
+    </head>
+    <body>
+    <table class="container">
+        ${generateEmailHeader()}
+        <tr>
+        <td class="content">
+            <h2>${i18n.translate('mail.subscriptionExpiration.title', lang, { status: statusMessage })}</h2>
+            <p>${i18n.translate('mail.subscriptionExpiration.greeting', lang)},</p>
+            <p>Pragniemy poinformować, że Twoja subskrypcja <strong>RunBotics</strong> ${statusMessage}</p>
```
bhszewczyk: missing translation

bhszewczyk (changes requested):

runbotics/runbotics-scheduler/src/mail/i18n.service.ts
Erquer: make lang optional with type SupportedLanguage, because then your logic of this function and checs if lang exists is wrong, because it always is present. 
now typesctipt only see 1 value for this function and there is no need of implify value if you use fallbackLang anyway.

runbotics/runbotics-scheduler/src/mail/i18n.service.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    translate(key: string, lang = 'en', vars: Record<string, string> = {}): string {
+        const fallbackLang: SupportedLanguage = 'en';
+        
+        const isValidLang = (lang: string): lang is SupportedLanguage => {
+            return lang === 'en' || lang === 'pl';
+        };
+
+        const effectiveLang: SupportedLanguage = (lang && isValidLang(lang) && this.translations[lang])
```
Erquer: here is no need of checking the lang, because it always have value

runbotics/runbotics-scheduler/src/mail/i18n.service.ts
Erquer: I wonder if there is not a better solution with enum. create enum with 2 values and then type SupportedLanguage = SuppLangEnum.ENGLISH | SuppLangEnum.POLISH (name the enum it diffrently) and instead of hardcoded strings, use enum in checks and fields

runbotics/runbotics-scheduler/src/mail/i18n.service.ts
```diff
@@ -0,0 +1,67 @@
...
+
+@Injectable()
+export class I18nService {
+    private logger = new Logger(I18nService.name);
+    private translations: Record<SupportedLanguage, TranslationObject>;
+
+    constructor() {
+        try {
+            this.translations = {
+                en: mailEn as TranslationObject,
```
Erquer: here you can use enum value for defining object { [SuppLangEnum.ENGLISH]: ... }

runbotics/runbotics-scheduler/src/mail/i18n.service.ts
Erquer: if you do, there is small chance of mistakes and it is more scalable, because if lang code change somehow, there is only 1 place to change

Erquer (changes requested):

runbotics/runbotics-scheduler/src/mail/i18n/en/mail.json
sniperdev: I added a body to the function

Erquer (approved):

bhszewczyk (approved):

PR #1165, opened 2025-10-14
[RPA-2500] add webhook migration files and definition of entities

Erquer:
## Types of changes

---


runbotics/runbotics-scheduler/src/webhook/entities/client-registration-webhook.entity.ts
```diff
@@ -0,0 +1,61 @@
...
+    @Column({ type: 'varchar' })
+    application_URL: string;
+
+    @Index()
+    @Column({ type: 'uuid', nullable: true })
+    authorization_id: string | null;
+
+    @Index()
+    @Column({ type: 'uuid', nullable: true })
+    client_registration_auth_id: string | null;
```
teawithsand: Why can't we use canonical TS naming convention with camelCase, like in [credential.entity.ts](https://github.com/runbotics/runbotics/blob/develop/runbotics/runbotics-scheduler/src/scheduler-database/credential/credential.entity.ts#L75)? Instead we could use `name` property in Column decorator in order to rename columns.

teawithsand (changes requested):

runbotics/runbotics-scheduler/src/webhook/entities/client-registration-webhook.entity.ts
Fiflon: +1

Erquer: copy paste from Figma, rest of column names in other tables are renamed to camelcase - its just a misspelling in this entity - i will fix it

runbotics/runbotics-scheduler/src/webhook/entities/webhook-payload.entity.ts
bhszewczyk, line 1: did we resign from action mapper? I thought we could add this and default it to "start_process". or this will be added later on if needed?

bhszewczyk (approved):

teawithsand (approved):

PR #1171, opened 2025-10-22
[RPA-2452] Implement language preference update functionality

sniperdev:
## Types of changes

---


runbotics/runbotics-scheduler/src/scheduler-database/account/account.controller.ts
bhszewczyk, line 1: We have user-account.controller.ts, so I think this logic should go there and logic from service should be in user.service.ts. If there is a reason you created whole module just for that - let me know, maybe I am missing something

bhszewczyk (changes requested):

bhszewczyk:
Also - delete this logic from java

Erquer (approved):
just look at the Change requests comments, they have a point

runbotics/runbotics-scheduler/src/scheduler-database/user/user.service.ts
```diff
@@ -363,4 +398,25 @@ export class UserService {
...
+        });
+
+        if (updateResult.affected === 0) {
+            throw new BadRequestException(
+                'User not found or language not updated'
+            );
+        }
+
+        return await this.userRepository.findOne({ where: { id: userId } });
+    }
```
bhszewczyk, line 419: I have 2 small issues with that, but happy to discuss:
1. you first check is for language support but I think more important is if that user even exists. if it doesn't exist why even bother to check language? why not use findOneBy (throw not found here if !user), then validate the language and then just save method (which returns entity - no need for findBy and update in separate calls). "Language not updated" is not needed info in my opinion - user wanted that language and that language is saved.
2. you throw BadRequestException but you say in message that not found (NotFoundException, 404) and not updated (304, but nestjs doesn't have that so imo this should not be an error or just HttpException('Not Modified', HttpStatus.NOT_MODIFIED)

bhszewczyk (changes requested):

bhszewczyk (approved):

PR #1185, opened 2025-11-07
[RPA-2523] Add process snapshot functionality

igorspychalaafo:
## Types of changes

---

### DIFF
--- a/runbotics/runbotics-scheduler/src/scheduler-database/process-snapshot/dto/process-snapshot-response.dto.ts
+++ b/runbotics/runbotics-scheduler/src/scheduler-database/process-snapshot/dto/process-snapshot-response.dto.ts
@@ -1,0 +1,40 @@
+import { ApiProperty } from '@nestjs/swagger';
+
+export class ProcessSnapshotResponseDto {
+    @ApiProperty({ example: 1 })
+    id: number;
+
+    @ApiProperty({ example: 1 })
+    processId: number;
+
+    @ApiProperty({ example: 'My Process' })
+    processName: string;
+
+    @ApiProperty({ example: 1 })
+    versionNumber: number;
+
+    @ApiProperty({ example: 'Initial Version' })
+    name: string;
+
+    @ApiProperty({
+        example: 'Snapshot created before major process changes'
+    })
+    description: string;
+
+    @ApiProperty({
+        example: 'admin@example.com',
+        description: 'Email of the user who created the snapshot'
+    })
+    createdByEmail: string;
+
+    @ApiProperty({ example: '2025-01-01T12:00:00.000Z' })
+    createdAt: string;
+}
+
+export class ProcessSnapshotDetailResponseDto extends ProcessSnapshotResponseDto {
+    @ApiProperty({
+        example: '<?xml version="1.0" encoding="UTF-8"?>...',
+        description: 'Decompressed process definition'
+    })
+    processDefinition: string;
+}
--- a/runbotics/runbotics-scheduler/src/scheduler-database/process-snapshot/process-snapshot.controller.ts
+++ b/runbotics/runbotics-scheduler/src/scheduler-database/process-snapshot/process-snapshot.controller.ts
@@ -1,0 +1,158 @@
+import {
+    Body,
+    Controller,
+    Get,
+    Logger,
+    Param,
+    ParseIntPipe,
+    Post,
+    UseInterceptors,
+} from '@nestjs/common';
+import {
+    ApiTags,
+    ApiOperation,
+    ApiCreatedResponse,
+    ApiOkResponse,
+    ApiNotFoundResponse,
+    ApiBadRequestResponse,
+    ApiParam,
+    ApiQuery,
+    ApiBody,
+} from '@nestjs/swagger';
+import { ProcessSnapshotService } from './process-snapshot.service';
+import { CreateSnapshotDto, CreateSnapshotSwaggerDto, createSnapshotSchema } from './dto/create-snapshot.dto';
+import { ProcessSnapshotResponseDto, ProcessSnapshotDetailResponseDto } from './dto/process-snapshot-response.dto';
+import { User as UserDecorator } from '#/utils/decorators/user.decorator';
+import { User } from '#/scheduler-database/user/user.entity';
+import { ZodValidationPipe } from '#/utils/pipes/zod-validation.pipe';
+import { Pageable, Paging } from '#/utils/page/pageable.decorator';
+import { Page } from '#/utils/page/page';
+import { TenantInterceptor } from '#/utils/interceptors/tenant.interceptor';
+import { FeatureKeys } from '#/auth/featureKey.decorator';
+import { FeatureKey } from 'runbotics-common';
+
+@ApiTags('Process Snapshots')
+@UseInterceptors(TenantInterceptor)
+@Controller('/api/scheduler/tenants/:tenantId/process-snapshots')
+export class ProcessSnapshotController {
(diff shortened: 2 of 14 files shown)


runbotics/runbotics-scheduler/src/scheduler-database/process-snapshot/process-snapshot.entity.ts
```diff
@@ -0,0 +1,62 @@
...
+import { ProcessEntity } from '#/scheduler-database/process/process.entity';
+import { User } from '#/scheduler-database/user/user.entity';
+import { ApiProperty } from '@nestjs/swagger';
+
+@Entity({ name: 'process_snapshot' })
+export class ProcessSnapshotEntity {
+    @ApiProperty({ example: 1 })
+    @PrimaryColumn({ type: 'bigint', transformer: numberTransformer })
+    @Generated()
+    id: number;
```
bhszewczyk, line 20: why not uuid?

runbotics/runbotics-scheduler/src/scheduler-database/process-snapshot/process-snapshot.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+import { ZodValidationPipe } from '#/utils/pipes/zod-validation.pipe';
+import { Pageable, Paging } from '#/utils/page/pageable.decorator';
+import { Page } from '#/utils/page/page';
+import { TenantInterceptor } from '#/utils/interceptors/tenant.interceptor';
+import { FeatureKeys } from '#/auth/featureKey.decorator';
+import { FeatureKey } from 'runbotics-common';
+
+@ApiTags('Process Snapshots')
+@UseInterceptors(TenantInterceptor)
+@Controller('/api/scheduler/tenants/:tenantId/process-snapshots')
```
bhszewczyk, line 36: why not tenants/:tenantID/processes/:processId/snapshots? and the for creating or getting snapshots you would not need anything else, you would not need to repeat that this is process-snapshots.../snapshots

runbotics/runbotics-scheduler/src/scheduler-database/process-snapshot/process-snapshot.service.ts
bhszewczyk, line 1: for all service methods: I don't see a check if the user who wants to do something with snapshots even has access to the process, user is only used here for logger

bhszewczyk (changes requested):

runbotics/runbotics-scheduler/src/scheduler-database/process-snapshot/dto/process-snapshot-response.dto.ts
```diff
@@ -0,0 +1,40 @@
+import { ApiProperty } from '@nestjs/swagger';
+
+export class ProcessSnapshotResponseDto {
```
Erquer, line 3: why not using zod to create dto's?

runbotics/runbotics-scheduler/src/scheduler-database/process-snapshot/process-snapshot.controller.ts
Erquer: good point - imo api should be under process resource if we want to stick to REST api.

runbotics/runbotics-scheduler/src/scheduler-database/process-snapshot/process-snapshot.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+    @Post('processes/:id/snapshots')
+    @FeatureKeys(FeatureKey.PROCESS_EDIT_INFO)
+    async createSnapshot(
+        @Param('tenantId') tenantId: string,
+        @Param('id', ParseIntPipe) processId: number,
+        @Body(new ZodValidationPipe(createSnapshotSchema)) createSnapshotDto: CreateSnapshotDto,
+        @UserDecorator() user: User,
+    ): Promise<ProcessSnapshotResponseDto> {
+        this.logger.log(`Creating snapshot for process ${processId}, tenant ${tenantId}`);
+        this.logger.log('jestem tutaj');
```
Erquer, line 68: ?? remove logs

runbotics/runbotics-scheduler/src/scheduler-database/process-snapshot/process-snapshot.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+        example: 1
+    })
+    @ApiBody({ type: CreateSnapshotSwaggerDto })
+    @ApiCreatedResponse({
+        description: 'Snapshot created successfully',
+        type: ProcessSnapshotResponseDto,
+    })
+    @ApiNotFoundResponse({ description: 'Process not found' })
+    @ApiBadRequestResponse({ description: 'Invalid input or empty process definition' })
+    @Post('processes/:id/snapshots')
```
Erquer, line 59: for example here: api call should look like /api/scheduler/tenants/tenantid/processes/processId/snapchosts

runbotics/runbotics-scheduler/src/scheduler-database/process-snapshot/process-snapshot.service.ts
```diff
@@ -0,0 +1,188 @@
...
+    ) {}
+
+    async createSnapshot(
+        processId: number,
+        userId: number,
+        createSnapshotDto: CreateSnapshotDto,
+    ): Promise<ProcessSnapshotResponseDto> {
+        this.logger.log(`Creating snapshot for process ${processId} by user ${userId}`);
+
+        const process = await this.processRepository.findOne({
```
Erquer, line 35: if you have so many database operations, you should do this as transaction, not as separate calls

Erquer: in this case there is only 1 dml query, so it can stay as it is, but in the future remember if there is more of it, use transactions

runbotics/runbotics-scheduler/src/scheduler-database/process-snapshot/process-snapshot.service.ts
```diff
@@ -0,0 +1,188 @@
...
+            const savedSnapshot = await this.processSnapshotRepository.save(snapshot);
+
+            const snapshotWithRelations = await this.processSnapshotRepository.findOne({
+                where: { id: savedSnapshot.id },
+                relations: ['process', 'createdBy'],
+            });
+
+            this.logger.log(`Created snapshot ${savedSnapshot.id} for process ${processId}`);
+
+            return this.mapToResponseDto(snapshotWithRelations);
```
Erquer, line 71: if you used zod as dto's you could just use dtoSchema.parse(entity) and return proper object

runbotics/runbotics-scheduler/src/scheduler-database/process-snapshot/process-snapshot.service.ts
```diff
@@ -0,0 +1,188 @@
...
+
+        if (!snapshot) {
+            throw new NotFoundException(`Snapshot with id ${snapshotId} not found`);
+        }
+
+        try {
+            const decompressedDefinition = await gunzipAsync(snapshot.processDefinition);
+            const processDefinition = decompressedDefinition.toString('utf8');
+
+            const responseDto = this.mapToResponseDto(snapshot) as ProcessSnapshotDetailResponseDto;
```
Erquer, line 123: you map to dto and still must cast as? something is wrong here

Erquer (changes requested):

PR #1186, opened 2025-11-12
[RPA-2532] Prevent scheduler from OOMing on a process using loop with 1000+ iterations

teawithsand:
Now messages are queued on scheduler side, rather than being processed all at once, which greatly decreases amount of memory that is required for allocating things like temp. variables for postgresql operations.

## Types of changes

---

### DIFF
--- a/runbotics-orchestrator/build.gradle
+++ b/runbotics-orchestrator/build.gradle
@@ -27,7 +27,7 @@ plugins {
 group = "com.runbotics"
 
 
-version = "3.6.0-SNAPSHOT.0"
+version = "3.6.0-SNAPSHOT.1"
 description = ""
 
 sourceCompatibility=11
--- a/runbotics/runbotics-scheduler/src/auth/guards/featureKey.guard.ts
+++ b/runbotics/runbotics-scheduler/src/auth/guards/featureKey.guard.ts
@@ -25,7 +25,9 @@ export class FeatureKeyGuard extends JwtAuthGuard {
         const request = context.switchToHttp().getRequest<AuthRequest>();
         const user = request.user;
 
-        return hasFeatureKeys(user, featureKeys);
+        const hasKeys = hasFeatureKeys(user, featureKeys);
+
+        return hasKeys;
     
     }
 }
--- a/runbotics/runbotics-scheduler/src/auth/guards/featureKey.guard.ts
+++ b/runbotics/runbotics-scheduler/src/auth/guards/featureKey.guard.ts
@@ -25,9 +25,7 @@ export class FeatureKeyGuard extends JwtAuthGuard {
         const request = context.switchToHttp().getRequest<AuthRequest>();
         const user = request.user;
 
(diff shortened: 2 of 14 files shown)


runbotics/runbotics-scheduler/src/auth/guards/featureKey.guard.ts
```diff
@@ -25,7 +25,9 @@ export class FeatureKeyGuard extends JwtAuthGuard {
         const request = context.switchToHttp().getRequest<AuthRequest>();
         const user = request.user;
 
-        return hasFeatureKeys(user, featureKeys);
+        const hasKeys = hasFeatureKeys(user, featureKeys);
```
Erquer: but why?

runbotics/runbotics-scheduler/src/auth/guards/featureKey.guard.ts
```diff
@@ -25,7 +25,9 @@ export class FeatureKeyGuard extends JwtAuthGuard {
...
+
+        return hasKeys;
```
bhszewczyk: I think this is unnecessary change

Erquer (approved):

runbotics-orchestrator/build.gradle
```diff
@@ -27,7 +27,7 @@ plugins {
 group = "com.runbotics"
 
 
-version = "3.6.0-SNAPSHOT.0"
+version = "3.6.0-SNAPSHOT.1"
```
bhszewczyk, line 30: since this should be included in release 3.5.0 DO NOT merge this to develop yet - let's discuss it on daily

runbotics/runbotics-scheduler/src/auth/guards/ws-bot-jwt.guard.ts
bhszewczyk, line 1: this will be another thing to test - add this to JIRA

bhszewczyk: now we will not log out validation errors (from auth.service validateParameterFromBot) - we do not want to miss this information if e.g. bot collection from UI is removed during the time the bot is connected. now bot will have token but it should not be connected anymore. the idea is good, as most of this logic is only needed for bot registration, but I would think about stashing these changes and add to the new version 3.6.0 rather as part of 3.5.0 - also to discuss

runbotics/runbotics-scheduler/src/utils/task-queue.spec.ts
```diff
@@ -0,0 +1,358 @@
+import { describe, it, expect, beforeEach } from 'vitest';
+import { TaskQueue } from './task-queue';
+
+describe('TaskQueue', () => {
+    let taskQueue: TaskQueue;
+
+    beforeEach(() => {
+        taskQueue = new TaskQueue(2);
```
bhszewczyk, line 8: I think this should be the same as DEFAULT_MAX_PARALLELISM

bhszewczyk (approved):

PR #1189, opened 2025-11-13
[RPA-2532] Scheduler OOM fix for release 3.5.0

teawithsand:
## Types of changes

---


runbotics/runbotics-scheduler/src/scheduler-database/tenant/tenant.service.ts
```diff
@@ -38,6 +38,16 @@ export class TenantService {
...
+    async isActive(tenantId: string): Promise<boolean> {
+        const result = await this.tenantRepository
+            .createQueryBuilder('tenant')
+            .select('tenant.active')
+            .where('tenant.id = :tenantId', { tenantId })
+            .getRawOne();
+        
+        return result?.active === true;
+    }
+
```
bhszewczyk: we have this logic in active-tenant.guard.ts so maybe you could extract this to be able to use it here and in guard as well? (active-tenant.guard.ts)

bhszewczyk (approved):

Erquer (approved):

